PR #6, opened 2018-03-27
Update regex to match https protocol

MikeSha:


gregnb:
Great find! Thanks mike

MikeSha:
You're welcome :)

PR #9, opened 2018-04-01
Safari issue with images if source is base64 data

MikeSha:
I've found that Safari doesn't execute onload if image has base64 data in src. As a result, target.print() will not be triggered.

I came up with a weird workaround, but it works for me. Looks like other browsers are ok with that :)


MikeSha:
@gregnb Yep. Looks like it works fine.

PR #10, opened 2018-04-02
Fix issue with old data.

MikeSha:
Sometimes "Printable" component get data that was previously loaded to the component. I mean, when you have a collection, and you need to update printable component with a new data.
 
I'm not sure about that solution, but I don't know how to catch something like componentWillReceiveProps outside the component. I believe there should be more acceptable solution, but that's all I have for now :)

Also I should thank you, @gregnb, for this amazing library. Looks like it's almost perfect :) At least, it does exactly what I need.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -23,14 +23,17 @@ class ReactToPrint extends React.Component {
 
   triggerPrint(target) {
     setTimeout(() => {
-
-      if (this.props.onBeforePrint) {
-        this.props.onBeforePrint();
-      }
-
       target.print();
       target.close();
     }, 500);
+  }
+
+  onClick = () => {
+    if (this.props.onBeforePrint) {
+      this.props.onBeforePrint();
+    }
+    /* We need to defer handlePrint in the callstack because it gives React time to rerender the DOM with a new data */
+    setTimeout(this.handlePrint, 1);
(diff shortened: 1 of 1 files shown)


gregnb:
Hey Mike,

I'm not sure I'm seeing the same thing happening. Take a look at the following:
https://wy2rl1nyzl.codesandbox.io/

Try sorting a list and then clicking print again.  We're going to have to find a better solution than more setTimeout's. Like you pointed out we don't know if it receives props, but then again in the demo I have above even a user changes the data in the table and clicks print again it seems to be working fine?

Thanks again for your help on these PR's to make this a very solid library

MikeSha:
@gregnb Sure, I totally agree that setTimeout is a bad solution.
I'll make a sandbox to reproduce the issue a little bit later.

MikeSha:
@gregnb I think it can be closed.

gregnb:
Sounds good! I'll do a release from yesterday's patch's sometime today

MikeSha:
Hi @gregnb . I promised you to create a sandbox with this problem, but I totally forgot about it. Anyway, I realized that this problem is not about react-to-print. It's about the architecture of the app :) So thank you, everything works as expected.

PR #20, opened 2018-06-07
Add bodyClass prop to pass class to printWindow body

mjoyce91:
An optional class can be passed to the `printWindow` body with the `bodyClass` prop.

Also updates the readme with new props.


gregnb:
Just curious what the use case was? Could you elaborate?

mjoyce91:
Sure - this was to provide a custom class to `body` so that I could target child classes within the print pop-up window. In my case, it was a requirement to be able to maintain the print window even after cancelling/printing the print job. Some of my styles were not formatted properly within the print window, so this way I can target and change them by providing `bodyClass="print-window"` and creating custom rules

```scss
.print-window {
  .class-to-be-fixed {
    ...
  }
}
```

gregnb:
Makes sense. Thanks for the writeup! I'll merge and be doing a release tonight

PR #24, opened 2018-06-20
Fixed: Uncaught DOMException: Failed to execute 'add' on 'DOMTokenList': The token provided must not be empty.

SudeepSidhu:
Turns out, if you have an empty class in `document.body.className` it'll cause an exception during `printWindow.document.body.classList.add`. We need to filter out those classnames before passing them on.


gregnb:
Nice thanks for discovering this issue @SudeepSidhu I will do a release later tonight after work

PR #25, opened 2018-06-22
Prop to pass options to window.open()

mjoyce91:
Adds a prop so that custom values can be passed to the third argument of `window.open()`. For example, if a user did not want to use the default value of `"status=no, toolbar=no, scrollbars=yes"` and wanted to use `"toolbar=yes, resizable=yes"`, they could pass an object which would then get mapped into a string with the same format. So, the following could done:

```jsx
<ReactToPrint
    {...otherProps}
    printWindowOptions: {
        toolbar: 'yes',
        resizable: 'yes',
    }
/>
```

And would be interpreted as `"toolbar=yes, resizable=yes"` to be passed in to `window.open()`.


gregnb:
Hey @mjoyce91 thanks for the PR, but I am going to decline because this package has been rewritten to use iframe's to fix cross origin issues.  It will no longer use new windows.

PR #27, opened 2018-06-27
add async func

CodeUndefind:
I added an asynchronous function. Please look at the code.


gregnb:
Hi @bianlin thanks for submitting the PR. I will unfortunately have to decline the PR since I am not looking at this moment to add async support. Thank you for your time

PR #30, opened 2018-07-03
Optional page margins

codesignist:
Page margins can change by user props


gregnb:
Sorry what I meant was make it so that a user an override all the styling. Not just the margins that way if other issues come in the future someone can just put exactly what they want

codesignist:
Overriding style is more clever way than applying only margin. I agre with you.
But I had a hurry, so I decided to make a pull request to solve the problem in short way.
You can delete this request. But we need style applying method as soon as possible.
Thank you.

PR #44, opened 2018-09-02
Support for IE11

SDShooter:
Hi, this should fix the IE11 issue.  when .close() was called a few lines down it was re-triggering the onload event in that browser.  Thanks for your help regarding the github process.


gregnb:
Thanks for the PR!

PR #50, opened 2018-10-03
Handle printing multiple refs

jalbatross:
Modified handlePrint to naively print an array of refs as opposed to a single ref when it receives an Array of refs.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -69,6 +69,11 @@ class ReactToPrint extends React.Component {
       return false;
     }
 
+    if (Array.isArray(contentEl) && contentEl.length > 0) {
+      this.handlePrintMultipleRefs(contentEl);
+      return;
+    }
+
     let printWindow = document.createElement('iframe');
     printWindow.style.position = 'absolute';
     printWindow.style.top = '-1000px';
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -182,4 +315,4 @@ class ReactToPrint extends React.Component {
 
 }
 
-export default ReactToPrint;
+export default ReactToPrint;
```
MatthewHerbst, line 318: Nit: it's generally a good idea to have a newline at the end of the file - players nicer with some parsing logic in some scripts.

src/index.js
```diff
@@ -171,6 +176,134 @@ class ReactToPrint extends React.Component {
 
   }
 
+  /**
+   * Handles printing for an array of references
+   * 
+   * @param  {ref[]} refArray 
+   * 
+   */
+  handlePrintMultipleRefs = (refArray) => {
```
MatthewHerbst, line 185: There's a huge amount of duplicate code between this function and the one above. Can the common parts be extracted into a shared helper?

MatthewHerbst:
Going to close this since #88 will make it obsolete - you could do a `.forEach` of everything you want to print and just call `.print` on each one

Issue #49, opened 2018-09-28
Unable to render in IE11 using React 15


davegw:
Hi there,

Thanks for the library, very useful! 

While doing some browser testing, I was unable to render the react-to-print component in IE11. I saw there was a similar issue (#41) that appeared to be resolved in 2.0.0-alpha-3 (the version I'm using). I am using react 15 in my project (package specifies react ^16.2), but I don't see anything specific in the code that would require use of react 16. Copying the code into my codebase as opposed to importing from the node package, fixes the problem, but just posting in case others have run into this issue.

brittonjg:
I may have run into the same issue as @davegw. If it helps, I can see the following entry reported in Sentry ```Object doesn't support property or method 'includes'```.

I am using React `16.5.2` and `2.0.0-alpha-3` of react-to-print.

Really appreciate your work, let me know if you need any further information.

Edit.

It looks like the issue is when the UserAgent is checked for whether IE 11 is being used or not here https://github.com/gregnb/react-to-print/blob/master/src/index.js#L95

Seems like includes is another thing not supported in IE 11 - https://caniuse.com/#search=includes

gregnb:
react-to-print@2.0.0-alpha-4 should solve this thanks to @MatthewHerbst

Issue #43, opened 2018-08-31
2.0.0-alpha-2 fails for IPhone+Safari/Chrome and Windows 10 + IE11


SDShooter:
Thanks for the library, but in it's current state it doesn't seem like it's ready for widespread use.  I guess the alpha version tag means you may already be aware there are issues.

Steps to reproduce:
1) Try to print using the provided react-to-print test page: https://kmmw7l39y7.codesandbox.io/#

IPhone + Safari Results: Safari shows blank page in print preview
IPhone + Chrome Results: Nothing happens when button is clicked.
Windows + IE11 Results: 'Hung/Long running script pop up', no print preview. 

Phone details: 
IPhone (Latest) Plus, IOS 11.2.6

Windows Details:
Windows 10 
IE11

gregnb:
I'll look into the iOS Safari issues but I regret to inform this library is not going to support IE11. There is support for EDGE

SDShooter:
As a developer myself, I totally understand not supporting ie11.  I can't
wait for ie to be a distant memory.

As someone with corporate overlords to answer to, unfortunately I can't
personally drop support for ie11 until there are no longer a significant
number of people using microsoft operating systems that lack Edge (so
use ie11 because they don't know any better).

gregnb:
I understand your pain as someone in your position. Unfortunately I only have so much time to invest. If you happen to find any clues in solving the issues or a PR to offer up I'd be more than happy to accept

SDShooter:
For sure, I understand time constraints and having to draw a line in the sand somewhere. 

If I find something useful for IE11, I'll let you know and/or submit a pull request.  However, I think the path of least resistance on my end is going to be to make a screen visible page that is formatted for print (all site specific headers and footers removed) so that the native browser printing works.

SDShooter:
Good news, I found a fix.  Will send a PR later today for your review.  Thanks again for the cool project. :-)

SDShooter:
I tried pushing a /bugfix/v2-ie11 branch, but received a permission error.  I think I need permission to do so.

gregnb:
@SDShooter You will first need to fork the repo. Make the changes there and then submit a PR to this repo

gregnb:
Published your fix in the latest version. If you have any other solutions I'd be happy to accept them

SDShooter:
[...]

PR #51, opened 2018-10-04
Remove use of `.includes` to better support IE

MatthewHerbst:
This fixes an issue identified in #49 that seems to have been first brought up in #43 

This replaces the one use of `Array.prototype.includes` with `Array.prototype.indexOf`, which can be used to achieve the same thing: `x.includes(y) === x.indexOf(y) > -1`. This allows the code to work in IE11 and earlier, which do not support `.includes`

I've also standardized the `browsers` targets across the `.babelrc` and `rollup.config.js` files to make them consistent, along with minor code formatting improvements.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -5,13 +5,13 @@
     "react",
     ["env", {
       "targets": {
-        "browsers": ["> 1%", "iOS >= 8", "Android >= 4"],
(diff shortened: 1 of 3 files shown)


gregnb:
Thanks for the PR!

PR #56, opened 2018-10-15
Solve safari print issue

joshkoo:
Solution for https://github.com/gregnb/react-to-print/issues/43

I am not really sure what is the exact purpose of windowRemove function, however ios devices have quite bit of problem when you call close or remove content before actually printing it.

Please test it before merging.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -37,18 +37,11 @@ class ReactToPrint extends React.Component {
     setTimeout(() => {
       target.contentWindow.focus();
       target.contentWindow.print();
-      this.removeWindow(target);
 
       if (onAfterPrint) {
         onAfterPrint();
       }
 
-    }, 500);
-  }
-
-  removeWindow(target) {
-    setTimeout(() => {
-      target.parentNode.removeChild(target);
     }, 500);
   }
 


gregnb:
So here's the issue we don't want newly added iframes to stay. We do want to clean those up and remove them but I don't know of any afterPrint event I can tie into to know that a user is done printing.

Any ideas?

joshkoo:
Then i would suggest adding a second setTimeout however if they haven't selected the printer, or try to switch later on, their content might get deleted.

To handle this properly, we need to find a way to check removal of print window from document maybe using something like window.matchMedia('print') doesn't exist after calling trigger print and at that point you can be sure that they have closed the print window and run your onAfterPrint and remove window function.

joshkoo:
Printing is a seriously not well implemented feature on all browsers, so even if you have a solution it is really hard to make it work properly on all browsers and devices.

As i couldn't find a proper easy print module, i have created my own that works on mobile and desktop safari and chrome by mixing couple of modules i found in npm including yours.

gregnb:
I think this could be the comprise. I really only want to remove the window so we don't leave a mess but what about just setting a focus listener that we remove after?

```
  triggerPrint(target) {
    const { onBeforePrint, onAfterPrint } = this.props;

    if (onBeforePrint) {
      onBeforePrint();
    }

    setTimeout(() => {
      target.contentWindow.focus();
      target.contentWindow.print();

      const removeFocusListener = () => {
        target.parentNode.removeChild(target);
        window.removeEventListener('focus', removeFocusListener);
      };

      window.addEventListener('focus', removeFocusListener);

      if (onAfterPrint) {
        onAfterPrint();
      }

    }, 500);
  }
```

ctriantaf:
Hello, I'm facing the same issue in Safari.
Any updates on this?

Edit: Tested the above code from @gregnb and everything looks good on the preview window (haven't actually print yet)!

joshkoo:
> I think this could be the comprise. I really only want to remove the window so we don't leave a mess but what about just setting a focus listener that we remove after?
> 
> ```
>   triggerPrint(target) {
>     const { onBeforePrint, onAfterPrint } = this.props;
> 
>     if (onBeforePrint) {
>       onBeforePrint();
>     }
> 
>     setTimeout(() => {
>       target.contentWindow.focus();
>       target.contentWindow.print();
> 
>       const removeFocusListener = () => {
>         target.parentNode.removeChild(target);
>         window.removeEventListener('focus', removeFocusListener);
>       };
> 
>       window.addEventListener('focus', removeFocusListener);
> 
>       if (onAfterPrint) {
>         onAfterPrint();
>       }
> 
>     }, 500);
>   }
> ```

I mean this would work however, even after focus you will need to keep the content until you close the window.

In this case if you haven't picked the printer or switch some settings safari will delete the content and you might print an empty paper.

What i would suggest is to keep the until you are done with printing all together, that is why on my app i delete the content once the printer is unmounting rather than after triggering print and making it not visible in body. 

If you have any other ideas, feel free to share and  i will update the code.

kodermax:
Can be to remove a previous iframe, and only after creating a new one?

image

joshkoo:
> Can be to remove a previous iframe, and only after creating a new one?
> 
> image

That is what i was doing locally, I have added visibility / display setting to hide the print media in normal view, still keeps it after printing and only delete it if the print content changes, someone click print again / recreates a frame and while unmounting.

My current contract is ending in a week, then i can take a look at this and push the latest fix.

Eli-Black-Work:
What abut both using [onafterprint](https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onafterprint), on browsers that support it, to remove the iframe, and also making it so that when one tries to print and there's already an existing iframe, the existing iframe first gets removed?

The idea being to have the iframe be removed cleanly on browsers that support `onafterprint` and limit it so that there's no more than one iframe on browsers that don't support `onafterprint`.

MatthewHerbst:
@joshkoo have you had any chances to look at this again? Thanks again for the PR and leading the discussion on this!

joshkoo:
Hi @MatthewHerbst, sorry i was away for a holiday. I will take a look at this tomorrow and push an update.

kodermax:
Hi @joshkoo. 
How are you doing with this?

kodermax:
I tested this PR, the problem is not solved

MatthewHerbst:
Hi @joshkoo are you still working on this? Thanks for your efforts so far!

eleonora-guidi:
So, I have the same problem on safari and really need it to work.
Following your ideas, I added a check before appending the printWindow to the document: if the printWindow is already present (from the previous time I tried to print something) I remove it.
Only then I proceed appending the new printWindow.

To do so I added an id to the iframe and proceeded to check if that's present. Do you have a better idea? Do you think this could work?

eleonora-guidi:
@MatthewHerbst can you check my previous comment, please? 
Would you like to see the code somewhere?

MatthewHerbst:
Hi @Ellenergic if you have a PR that is different from this one, please of course show it to us :) Thanks!

eleonora-guidi:
> Hi @Ellenergic if you have a PR that is different from this one, please of course show it to us :) Thanks!

HI! Ok @MatthewHerbst, I'll open a new PR with the updated code, let me know what you think about that.  See the new PR #134

MatthewHerbst:
This should be solved as of #134. Will be publishing a new version shortly.

MatthewHerbst:
v2.1.3 is published! Please let me know if you run into an other issues

PR #59, opened 2018-10-29
Upgrade to babel-7

MatthewHerbst:
This upgrades babel to version 7. It also upgrades the various plugins that we use to their version 7 versions. Of note, the `rollup-plugin-babel` was upgraded to its newest version which uses babel-7 as well. As part of that upgrade, I changed the rollup build config to just use the `.babelrc` located in the project root, rather than maintaining a duplicate copy of it in the rollup config. Further, rollup now provides the babel `external-helpers` functionality, so we do not need to include it ourselves in our babel config, meaning it can be removed from this project.

I initially ran [`npx babel-upgrade`](https://github.com/babel/babel-upgrade) against the repo and then manually made the changes (since that script doesn't modify `package-lock.json`). For the rollup changes, I simply followed their [babel docs](https://rollupjs.org/guide/en#babel), which have been updated to show babel-7 examples. Note that I did not create a separate `.babelrc` file since we have no tests in this repo.

I tested this works by running all of the scripts (`npm run build`, `npm run webpack`, `npm run webpack-dev-server`). I also opened the page from the `npm run webpack-dev-server` and successfully "printed" the contents of the page.


gregnb:
Fantastic PR. Thanks for doing this clean up work

PR #61, opened 2018-10-29
Handle empty href attributes

### LINKED ISSUE #58: Fails to load print window when a malformed(?) link is on page
@gregnb 
I hit this error because I have a browser extension that adds the following css to every page of mine:
```
<link type="text/css" id="dark-mode" rel="stylesheet" href="">
```

Here's a codesandbox where I manually add that css to the sandbox page and as you can see it fails to load the print window: :  https://codesandbox.io/s/4jn3y155z7

I've tracked down what's happening: 
Ideally the link's onload or onerror would be triggered which would call `markLoaded` (https://github.com/gregnb/react-to-print/blob/master/src/index.js#L152): 
```js 
            newHeadEl.onload = markLoaded.bind(null, 'link');
            newHeadEl.onerror = markLoaded.bind(null, 'link');
```

Unfortunately neither onload or onerror is being called and thus `markLoaded` is never triggered. I'm not quite sure why onload/onerror isn't working in this case. Probably because the link doesn't actually link to an href? Messing around with the link 

Any thoughts on how to safeguard against <link> elements causing this to occur? 

Maybe the link logic should be disabled by default until this is fixed? 

Thanks!

MatthewHerbst:
This fixes #58 

`<link>` tags with empty `href` attributes are [INVALID HTML](https://www.w3.org/TR/html50/document-metadata.html#attr-link-href). Further, browsers often do [unexpected things](https://gtmetrix.com/avoid-empty-src-or-href.html) when they encounter an empty `href`. For example, some will attempt to load the current page. Some will event attempt to load the current page's parent directory. (My current browser, Chrome 70, attempts to load the current page.)

To solve these problems, this PR causes `react-to-print` to skip loading `<link>` tags when they have an empty `href` attribute.

This PR also actually makes use of the `markLoaded` method's parameters (I changed them up a bit), and it removes some unneeded whitespace.

Testing strategy:

1. Open the test page at `http://0.0.0.0:5060/#` after running `npm run webpack-dev-server`
2. Test that everything works normally (there are no malformed links on the test page)
3. In console, do:
    1. `h = document.getElementsByTagName('head')`
    2. `h[0].innerHTML = h[0].innerHTML + '<link type="text/css" id="dark-mode" rel="stylesheet" href="">';`
4. Click "Print this out!"

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 [![dependencies Status](https://david-dm.org/gregnb/react-to-print/status.svg)](https://david-dm.org/gregnb/react-to-print)
 [![npm version](https://badge.fury.io/js/react-to-print.svg)](https://badge.fury.io/js/react-to-print)
 
-So you've created a React component but would love to give end users the ability to print out the contents of that component. This package aims to solve that by popping up a new print window with CSS styles copied over as well.  
+So you've created a React component but would love to give end users the ability to print out the contents of that component. This package aims to solve that by popping up a new print window with CSS styles copied over as well.
 
 ## Install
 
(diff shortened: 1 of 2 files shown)


gregnb:
Awesome! Thanks again for another great PR!

PR #65, opened 2018-11-01
Addition cleanup

MatthewHerbst:
This performs some additional cleanup:

1) Upgrades all non-breaking dependencies to their latest versions
2) it upgrades ESLint to the latest version and actually uses it - before we had no `.eslintrc` so for most people no linting was happening. I've extended the well-known [AirBnB style guide](https://github.com/airbnb/javascript). Let me know if you'd like to use a different rule set instead
3) Move `react-dom` to `peerDependencies`: makes sense since `react` is there too :)
4) Add `react` and `react-dom` to `devDependencies` for simpler dev experience (plus manually installing them for local dev purposes, even without the `--save` flag, causes changes in `package-lock.json` which is undesirable since they were not requested.

As part of the ESLint upgrade I also had to fix all the linting issues with the lib, or turn off rules as needed where they didn't make sense for us, or in one instance, where I couldn't get it working. Part of fixing the linting issues meant extracting one of the components in the example page since the linter wisely only wants one React component declared per-file.


gregnb:
Looks great to me just revert the .js -> .jsx change. Rather it stay how it was 

Thanks!

MatthewHerbst:
Hi @gregnb thanks for the review. Any chance I could convince you on the `.jsx`? It's pretty heavily pushed by the [AirBnB React Style Guide](https://github.com/airbnb/javascript/tree/master/react#naming) and others. I also find it helpful for developer tooling, since some tools won't parse a file as JSX unless it's `.jsx`.

gregnb:
Sorry man, something I want to stay away from. Be glad to accept this PR once that is changed

MatthewHerbst:
@gregnb done. Thanks for being responsive to PRs 👍

gregnb:
No problem. Thanks for your help and understanding!

PR #72, opened 2018-11-27
Don't include /example in published npm package

MatthewHerbst:
`/example` is code for development and is not needed by end users. This saves users from having to download code they will not use.


gregnb:
Agreed!

PR #73, opened 2018-11-27
Remove `.npmignore`

MatthewHerbst:
This reverts #72 

I didn't realize this library already specifies the `files` attribute in its `package.json` file, which acts as a whitelist for `npm pack/publish` so we don't need to create a blacklist with a `.npmignore` file.

I made the initial commit in #72 because I had seen a bug post about it to another package I manage and for some reason made the change here instead of there - SORRY!

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,1 +1,0 @@
-example/


MatthewHerbst:
Assuming you are on npm v6, you can see which files will be included in the package by running `npm pack && tar -xvzf *.tgz && rm -rf package *.tgz`. I clearly should have done that before creating #72, sorry again.

gregnb:
No worries man. Merged

Issue #68, opened 2018-11-09
Export a method to trigger print


leeseean:
I want to print by Api instead of click,
just like ``` target.print() ```

MatthewHerbst:
Do you mean without using the React component at all? Right now you can do:

```javascript
render() {
  componentDidMount() {
    this.ref.handlePrint();
  }

  return (
        <ReactToPrint
          ...
          ref={ref => this.ref = ref}
        />
  );
}
```

I'd be interested in moving the print API outside of the component itself. How do you imagine setting up `target` in your example?

leeseean:
> Do you mean without using the React component at all? Right now you can do:
> 
> ```js
> render() {
>   componentDidMount() {
>     this.ref.handlePrint();
>   }
> 
>   return (
>         <ReactToPrint
>           ...
>           ref={ref => this.ref = ref}
>         />
>   );
> }
> ```
> 
> I'd be interested in moving the print API outside of the component itself. How do you imagine setting up `target` in your example?

target is a ref or domNode

mengbimao:
东西是好东西,但是貌似我这个项目用不了,我照着上面来做了,但是样式貌似没有打印出来,无论怎么说还是辛苦啦,点一个星,不成敬意

mengbimao:
![image](https://user-images.githubusercontent.com/29363520/49124648-2d435a00-f2f8-11e8-8887-f3456fc1b26f.png)
![image](https://user-images.githubusercontent.com/29363520/49124668-3fbd9380-f2f8-11e8-9a21-6ec2499baec2.png)
![image](https://user-images.githubusercontent.com/29363520/49124687-54019080-f2f8-11e8-9314-e45142689e56.png)

Eli-Black-Work:
Thanks for the info :)  It would be great if this were documented, though.

Also, any chance of making the trigger() function optional? :)

yefriddavid:
For me works! 

this

  manualPrint() {

    this.printRef.handlePrint();

  }

render(){
        <ReactToPrint
          ref={el => (this.printRef = el)}
          trigger={() => <a href="#" id="david">Imprimir</a>}
          content={() => this.componentRef}
        />

......

heyjaypray:
I'm currently having a similar issue.  I'm currently using material-table and would like to have a button that you click that will print the info contained passed into a component.

[...]

PR #77, opened 2018-12-15
Trigger print via state change

steeevio:
I was struggling to put the `<ReactToPrint>` within sibling / child components to the `<ComponentToPrint>` component. 

Maybe there is a better way of doing this but for me it seemed to make it more flexible if I could trigger print via state. then I can do it with a button from anywhere.

I just added a some of props `<ReactToPrint>` that allow the user to decide if they want to trigger from state
`useState={true}`

check the current print state - set to true to run print
`statePrint= {true}`

user should add a function that sets the print state back to false after use.
`handleStatePrintOff={this.handleStatePrintOff} `

### DIFF
--- a/example/index.js
+++ b/example/index.js
@@ -9,7 +9,7 @@ class Example extends React.Component {
   constructor(props){
     super(props);
     this.state = {
-      print:false,
+      print: false,
   }
       
   handleAfterPrint = () => {
@@ -32,11 +32,11 @@ class Example extends React.Component {
     this.componentRef = ref;
   }
 
-  handleStatePrint = (e)=>{
-    this.setState({print : true});
+  handleStatePrint = (e) => {
+    this.setState({ print: true });
   }
-  handleStatePrintOff = (e)=>{
-    this.setState({print : false});
+  handleStatePrintOff = (e) => {
+    this.setState({ print: false });
   }
 
 
@@ -52,7 +52,6 @@ class Example extends React.Component {
           onAfterPrint={this.handleAfterPrint}
           statePrint={this.state.print}
           handleStatePrintOff={this.handleStatePrintOff}
-          useState={false}
         />
         <ComponentToPrint ref={this.setRef} />
       </div>
--- a/example/index.js
+++ b/example/index.js
@@ -54,6 +54,7 @@ class Example extends React.Component {
           onAfterPrint={this.handleAfterPrint}
           statePrint={this.state.print}
           handleStatePrintOff={this.handleStatePrintOff}
+          useState={false}
         />
         <ComponentToPrint ref={this.setRef} />
       </div>
--- a/example/index.js
+++ b/example/index.js
@@ -34,11 +34,9 @@ class Example extends React.Component {
 
   handleStatePrint = (e)=>{
     this.setState({print : true});
-    console.log("print-r go")
   }
   handleStatePrintOff = (e)=>{
     this.setState({print : false});
-    console.log("print-r stop")
   }
 
 
--- a/example/index.js
+++ b/example/index.js
@@ -5,6 +5,13 @@ import ReactToPrint from '../src';
 import ComponentToPrint from './ComponentToPrint';
 
 class Example extends React.Component {
+  
+  constructor(props){
+    super(props);
+    this.state = {
+      print:false,
+  }
+      
   handleAfterPrint = () => {
     console.log('after print!'); // eslint-disable-line no-console
   }
@@ -25,14 +32,28 @@ class Example extends React.Component {
     this.componentRef = ref;
   }
 
+  handleStatePrint = (e)=>{
+    this.setState({print : true});
+    console.log("print-r go")
+  }
+  handleStatePrintOff = (e)=>{
+    this.setState({print : false});
+    console.log("print-r stop")
+  }
+
+
+
   render() {
     return (
       <div>
+        <button onClick={this.handleStatePrint} type="button">Print with state!</button>
         <ReactToPrint
           trigger={this.renderTrigger}
           content={this.renderContent}
           onBeforePrint={this.handleBeforePrint}
           onAfterPrint={this.handleAfterPrint}
+          statePrint={this.state.print}
+          handleStatePrintOff={this.handleStatePrintOff}
         />
         <ComponentToPrint ref={this.setRef} />
       </div>
--- a/src/index.js
+++ b/src/index.js
@@ -178,7 +178,7 @@ class ReactToPrint extends React.Component {
       }
     };
 
-    console.log("print-r stop now")
+    
     this.props.handleStatePrintOff;
 
     document.body.appendChild(printWindow);
@@ -196,7 +196,7 @@ class ReactToPrint extends React.Component {
     if(this.props.useState === true){
       
       if(this.props.statePrint === true){
-        console.log("print-r go now")
+       
         this.handlePrint()
       } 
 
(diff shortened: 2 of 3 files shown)


example/index.js
```diff
@@ -5,6 +5,13 @@ import ReactToPrint from '../src';
 import ComponentToPrint from './ComponentToPrint';
 
 class Example extends React.Component {
+  
+  constructor(props){
+    super(props);
+    this.state = {
+      print:false,
```
MatthewHerbst: Nit:
```suggestion
      print: false,
```

example/index.js
```diff
@@ -25,14 +32,27 @@ class Example extends React.Component {
     this.componentRef = ref;
   }
 
+  handleStatePrint = (e)=>{
```
MatthewHerbst: Nit:

```suggestion
  handleStatePrint = (e) => {
```

example/index.js
```diff
@@ -25,14 +32,27 @@ class Example extends React.Component {
...
+    this.setState({print : true});
+  }
+  handleStatePrintOff = (e)=>{
```
MatthewHerbst: Nit:
```suggestion
  handleStatePrintOff = (e) => {
```

example/index.js
```diff
@@ -25,14 +32,27 @@ class Example extends React.Component {
...
+    this.setState({print : false});
```
MatthewHerbst: Nit:
```suggestion
    this.setState({ print: false });
```

example/index.js
```diff
@@ -25,14 +32,27 @@ class Example extends React.Component {
     this.componentRef = ref;
   }
 
+  handleStatePrint = (e)=>{
+    this.setState({print : true});
```
MatthewHerbst: Nit:
```suggestion
    this.setState({ print: true });
```

example/index.js
```diff
@@ -25,14 +32,27 @@ class Example extends React.Component {
...
       <div>
+        <button onClick={this.handleStatePrint} type="button">Print with state!</button>
         <ReactToPrint
           trigger={this.renderTrigger}
           content={this.renderContent}
           onBeforePrint={this.handleBeforePrint}
           onAfterPrint={this.handleAfterPrint}
+          statePrint={this.state.print}
+          handleStatePrintOff={this.handleStatePrintOff}
+          useState={false}
```
MatthewHerbst: The default value of `useState` should be `false`, so you should never need to pass the prop with `false`

src/index.js
```diff
@@ -53,6 +53,7 @@ class ReactToPrint extends React.Component {
   }
 
   handlePrint = () => {
+
```
MatthewHerbst: Nit:
```suggestion
```

src/index.js
```diff
@@ -177,6 +178,9 @@ class ReactToPrint extends React.Component {
       }
     };
 
+
```
MatthewHerbst: Nit: please remove these extra newlines

src/index.js
```diff
@@ -189,10 +193,31 @@ class ReactToPrint extends React.Component {
...
-      ref: this.setRef,
-    });
+    if(this.props.useState === true){
+      
+      if(this.props.statePrint === true){
+        this.handlePrint()
+      } 
+
+      return (
+        <div></div>
```
MatthewHerbst: Please use `<div />`. Also, please make sure to add a semicolon after the closing `)`

src/index.js
```diff
@@ -189,10 +193,31 @@ class ReactToPrint extends React.Component {
       trigger,
     } = this.props;
 
-    return React.cloneElement(trigger(), {
-      onClick: this.handlePrint,
-      ref: this.setRef,
-    });
+    if(this.props.useState === true){
+
```
MatthewHerbst:
```suggestion
```

src/index.js
```diff
@@ -189,10 +193,31 @@ class ReactToPrint extends React.Component {
...
+    if(this.props.useState === true){
+      
+      if(this.props.statePrint === true){
+        this.handlePrint()
+      } 
+
+      return (
+        <div></div>
+      )     
+    }else{
```
MatthewHerbst:
```suggestion
    } else {
```

src/index.js
MatthewHerbst: Also, why did you choose to return an empty `div` here rather than `null`?

src/index.js
```diff
@@ -189,10 +193,31 @@ class ReactToPrint extends React.Component {
...
+      
+      return React.cloneElement(trigger(), {
+        onClick: this.handlePrint, 
+        ref: this.setRef,
+      });
+    }
+
```
MatthewHerbst: Nit: please remove extra newlines

src/index.js
```diff
@@ -189,10 +193,31 @@ class ReactToPrint extends React.Component {
       trigger,
     } = this.props;
 
-    return React.cloneElement(trigger(), {
-      onClick: this.handlePrint,
-      ref: this.setRef,
-    });
+    if(this.props.useState === true){
```
MatthewHerbst: Please destructure this prop above

src/index.js
```diff
@@ -189,10 +193,31 @@ class ReactToPrint extends React.Component {
...
+      
+      if(this.props.statePrint === true){
```
MatthewHerbst: Please destructure this prop above

MatthewHerbst (changes requested):
Thanks for the PR! A few quick things first:

1. I left a bunch of style related comments on the PR that would be great if you could clean up
2. There shouldn't have been any changes to `package-lock.json` with this PR. You likely did an `npm install` instead of `npm ci`. Please do not commit an updated `package-lock.json` with this PR
3. Please add the new props to the `propTypes` definitions

I think overall what you are really trying to do is solve #68, which wants a way to print without having to use the button. While I think this approach is a good bandaid attempt to that before the work for #68 is complete, the issue is that if/when we do the work for #68, removing these changes would be a breaking change which should ideally be avoided. 

Would you be interested in working on #68 together rather than attempting this approach?

example/index.js
steeevio: Hi, this was my first stab at a pull request, so thanks for taking the time over the feedback! I've a had a go at changing it as required - hopefully I got everything, if not let me know.

I'm new to react so not really sure how much help I can be in dealing with #68, I'm happy to try but honestly you might just end up spending more time explaining things to me.

MatthewHerbst: Pretty great stab for your first PR, well done!

I actually think #68 doesn't have much to do with React - the goal is to take code that's currently baked into the component and move it into its own standalone JS. Then the React component can depend on it, and we can make it available to others. For example, a sample export might be:

```js
import { print } from 'react-to-print'
```

Then you could just call `.print(<someReactRef>|<someDOMNode>)`

That would solve what you are trying to do here, which is control when you print via external code. I'll take another look at it to see what a good path forward should be.

Do note I'm not the package owner, just an active contributor.

gregnb:
hey @MatthewHerbst really appreciate all your help. I've added push access for you

MatthewHerbst:
@gregnb thank you! Looking forward to getting some free time to address these issues :)

example/index.js
steeevio: Thanks! I see what you mean, yes ultimately that seems like a better solution.

MatthewHerbst:
Going to close this in favor of #88 

@steeevio would love any feedback you have about that PR, which is still in early stages though the meat of it is done

PR #80, opened 2018-12-21
Add TypeScript declaration

### LINKED ISSUE #79: Typescript definitions
It would be nice to have TypeScript definitions

cmckni3:
Fixes #79


cmckni3:
In case you're curious, I generated the declaration file with the following command.

```
tsc --allowSyntheticDefaultImports --downlevelIteration --lib es6,dom,dom.iterable --declaration src/index.ts
```

cmckni3:
`src/index.ts` is a fixed up version with types based on `src/index.js`.

See https://github.com/cmckni3/react-to-print/blob/typescript/src/index.ts

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "react-to-print",
-  "version": "2.0.0-alpha-4",
+  "version": "2.0.0-alpha-5",
```
MatthewHerbst, line 3: Nit: we should bump the version in a separate commit

MatthewHerbst (approved):
Looks great, thank you!!!

Only one small ask: could you please add a script to the `scripts` section of `package.json` that would do the `tsc --allowSyntheticDefaultImports --downlevelIteration --lib es6,dom,dom.iterable --declaration src/index.ts` command? Adding `tsc` as a dev-dependency of the package is totally fine if needed. That way others can be sure to keep the file up-to-date if they change the API at all.

package-lock.json
cmckni3: This is due to the package version being bumped for release but not running `npm install` after. I can remove that change though from when I ran `npm install`

Yet another one of my gripes about the npm lockfile. ¯\_(ツ)_/¯

MatthewHerbst: Oh, I totally thought this was in `package.json` heh. Seems like we should do this, so just leave it in 👍

cmckni3:
Sure, I can add that as a script. I wanted approval before introducing TypeScript as a dependency. I will copy over the type information as jsdoc comments as well instead of changing over to a .ts file.

cmckni3:
Looks like it will have to go another direction. The declaration files can't be generated from JS files yet and there's no way to specify TS types in JS files and have it be valid.

Maybe babel/rollup could help but I believe that should be a future PR.

cmckni3:
I ended up splitting out the jsdoc changes to expedite the process.

Unfortunate that there is no easy to way to generate a TS declaration file from a JavaScript file.

MatthewHerbst:
Thanks for making the changes! Would be happy to see more TS related commits/PRs 💯

cmckni3:
Definitely. I would get the other open PRs merged before switching over to T’s completely.

PR #81, opened 2018-12-21
fix(props): Change defaultProps callback values

cmckni3:
* Change onAfterPrint and onBeforePrint values in defaultProps to undefined

Required for #79


MatthewHerbst (approved):
Thanks!

PR #87, opened 2019-01-04
hungnt167/react-to-print#7597

hungnt167:
merge me or try another solution.
This is result: https://goo.gl/Bv31Kq

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -34,6 +34,8 @@ class ReactToPrint extends React.Component {
     }, 500);
   }
 
+  currentPrintWindow;
+
   triggerPrint = (target) => {
     const { onBeforePrint, onAfterPrint } = this.props;
 
@@ -44,7 +46,6 @@ class ReactToPrint extends React.Component {
     setTimeout(() => {
       target.contentWindow.focus();
       target.contentWindow.print();
-      this.removeWindow(target);
 
       if (onAfterPrint) {
         onAfterPrint();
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -67,6 +68,9 @@ class ReactToPrint extends React.Component {
       return;
     }
 
+    /** clear prev print window in here */
+    this.currentPrintWindow && this.removeWindow(this.currentPrintWindow);
```
MatthewHerbst, line 72: How does this work? `currentPrintWindow` is never assigned to anything, so this will never actually remove the previous print window

MatthewHerbst (changes requested):
You didn't assign `currentPrintWindow` anywhere

src/index.js
```diff
@@ -67,7 +68,11 @@ class ReactToPrint extends React.Component {
...
```
hungnt167, line 72: Sorry @MatthewHerbst , i had missing code commit,  I just have updated https://github.com/gregnb/react-to-print/commit/16d65a0752721541d1143e9f80d5a99a68070641, Please review again or add new an alternative solution.

MatthewHerbst, line 72: This will now run before we've actually printed anything, resulting in a blank page being printed.

MatthewHerbst (changes requested):
Have you looked at #56 at all? I think you and him are trying to solve the same issue

PR #88, opened 2019-01-04
WIP [DO NOT MERGE]: Extract core print logic

### LINKED ISSUE #68: Export a method to trigger print

MatthewHerbst:
WIP: this is still being worked on, but wanted to post it to show what progress has been made and to allow comments. So far I have the current print example working without issues!

This extracts the core print logic away from the React component and makes it available as plain JS. This allows us the ability to export the functionality so that end users can use it without having to actually use it in a JSX context - they can now print via the API.

This will close #68 and #77

- [x] Extract core logic
- [ ] Add tests
- [ ] Add API printing example
- [ ] Update docs
- [ ] Attempt to run through basic use cases to ensure no regressions (which would imply the need for additional tests)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -5,8 +5,16 @@ module.exports = {
         "browser": true,
     },
     "rules": {
+        "arrow-body-style": ["off"],
+        "consistent-return": ["off"],
         "import/import/no-unresolved": ["off"],
+        "max-len": ["error", {
+            "code": 100,
+            "ignoreComments": true,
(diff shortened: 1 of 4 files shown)


gregnb:
@MatthewHerbst Looks very promising! Often times I end up solving this issue by providing an callback to set a reference to the instance method but this will be better

```
<ReactToPrint
  triggerAction={fn => this.triggerFn = fn}
 ...
/>
```

PR #91, opened 2019-01-16
Fix imports and exports for typescript

alexdevero:
`react-to-print` is causing failing build in project using TypeScript. The reason are 1) incorrect imports for `React` and `PropTypes` and 2) `ReactToPrint` not exported as default.

Errors:
```
Failed to compile.

C:\temp\node_modules\react-to-print\lib\index.d.ts
(1,8): error TS1192: Module '"C:/temp/node_modules/@types/react/index"' has no default export.
```

```
Failed to compile.

C:\temp\node_modules\react-to-print\lib\index.d.ts
(2,8): error TS1192: Module '"C:/temp/node_modules/@types/prop-types/index"' has no default export.
```

```
Failed to compile.

./src/Foobar.tsx
(14,8): error TS1192: Module '"C:/temp/node_modules/react-to-print/lib/index"' has no default export.
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,10 +1,10 @@
-import React from 'react';
-import PropTypes from 'prop-types';
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
 
 /*~ This declaration specifies that the class constructor function
  *~ is the exported object from the file
  */
-export = ReactToPrint;
+export default ReactToPrint;
 
 declare class ReactToPrint extends React.Component<ReactToPrint.ReactToPrintProps> {
   static propTypes: {


index.d.ts
```diff
@@ -1,10 +1,10 @@
-import React from 'react';
-import PropTypes from 'prop-types';
+import * as React from 'react';
```
MatthewHerbst, line 1: Can you please explain why this format is needed over the current format? Is this something that's specific to TypeScript?

MatthewHerbst (commented):
Thanks for the PR! I left one quick question to address please

index.d.ts
alexdevero: This is specific to TypeScript and ES6 - https://github.com/Microsoft/TypeScript/issues/3337#issuecomment-107971371. The `import React from "react";` works with babel and other transpilers. It requires having specific settings in tsconfig, which may not be set. These settings are not set as default. When these settings are not set, TypeScript will fail.

MatthewHerbst: Awesome, thanks for the link

alexdevero: Thank you for accepting my PR @MatthewHerbst!

PR #97, opened 2019-02-20
README updated

andydev404:
Calling component to print from functional components with hooks

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ class Example extends React.Component {
 }
 ```
 
-### Calling from functional components with hooks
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -70,27 +70,75 @@ class Example extends React.Component {
     );
   }
 }
+```
 
+### Calling from functional components with hooks
```
MatthewHerbst:
```suggestion
### Calling from functional components with [hooks](https://reactjs.org/docs/hooks-intro.html)
```

MatthewHerbst (approved):
Thanks! Mind adding that link in just so people can easily know what hooks are since it's a pretty new concept? Want to make sure people not on React ^16.8 don't try to use it

PR #98, opened 2019-02-20
Catch 'member not found' error on cssText

### LINKED ISSUE #75: Script Error "Member Not Found" in IE11
I am using the latest version of react-to-print, and I couldn't make it to show the print dialog in IE 11 browser. It has the following error in the console `SCRIPT3: Member not found.` Clicking inside for details, it points to the line

```
var p = e.querySelectorAll("canvas");
if ([].concat(toConsumableArray(p)).forEach(function (e, t) {
    e.getContext("2d").drawImage(n[t], 0, 0);
}), !1 !== o) {
    var d = document.querySelectorAll('style, link[rel="stylesheet"]');
    [].concat(toConsumableArray(d)).forEach(function (t, n) {
        var r = e.createElement(t.tagName), o = "";
        if ("STYLE" === t.tagName) {
            if (t.sheet) {
                for (var a = 0; a < t.sheet.cssRules.length; a++)
==>                   o += t.sheet.cssRules[a].cssText + "\r\n";
                r.setAttribute("id", "react-to-print-" + n), r.appendChild(e.createTextNode(o));
            }
        }
        else {
            [].concat(toConsumableArray(t.attributes)).forEach(function (e) {
                r.setAttribute(e.nodeName, e.nodeValue);
            }), r.onload = u.bind(null, "link"), r.onerror = u.bind(null, "link");
        }
        e.head.appendChild(r);
    });
}
```
It works fine in Chrome and Firefox, I am not sure how to proceed now, does anyone have the same issue? Thanks in advance!

OmgBrown:
Issue found in IE11 when css/style has 'transform' property. Fixes #75


src/index.js
```diff
@@ -134,7 +134,10 @@ class ReactToPrint extends React.Component {
 
             if (node.sheet) {
               for (let i = 0; i < node.sheet.cssRules.length; i++) {
-                styleCSS += node.sheet.cssRules[i].cssText + "\r\n";
+                //catch 'member not found' error on cssText
+                if (typeof node.sheet.cssRules[i].cssText === "string") {
```
MatthewHerbst, line 150: Do we need to check that it's a string (can it be anything else?), or just that it exists?

MatthewHerbst (approved):
My only concern is that this functionality does something, and that this change, rather than making the functionality work for IE, simply says we won't do this on IE. Will this lead to inconsistent results across browsers for example, where it prints one way in Chrome and another way in IE?

src/index.js
OmgBrown: Null test didn't work. When it's a string value, it is an acceptable styleCSS. It worked fine with chrome and FF.

MatthewHerbst: `if(node.sheet.cssRules[i].cssText)` didn't work?

Out of curiosity, what does IE use instead of this?

OmgBrown: `if(node.sheet.cssRules[i].cssText)` this catch didn't work. 

In IE11, cssText has an UNKNOWN value when it sees a CSSKeyframesRule with 'transform' property but not a STRING. Chrome and FF has the correct STRING value.

[https://stackoverflow.com/questions/10982739/typeof-returning-unknown-in-ie](https://stackoverflow.com/questions/10982739/typeof-returning-unknown-in-ie )
 _"Basically that’s MS’s answer if you try to test or access something that’s not a true part of their JScript engine."_

MatthewHerbst:
Happy to merge once the conflicts are resolved - looks like just a bad rebase maybe?

MatthewHerbst:
@OmgBrown thanks. One minor nit: any chance you could squash the 4 commits down into 1 please?

You can do this by running (assuming in a linux-style terminal):

1. `git rebase -i HEAD~4`
2. Go into interactive mode in the shell by pressing the `I` key on your keyboard
3. For the bottom three commits (the three most recent ones), change them from `pick` to `squash`
4. Exit interactive mode the rebase by hitting your `ESC`
5. Save the rebase: type `:wq` and then press enter

OmgBrown:
I created a new fork and added my change. can we close this pull request and work on a clean one?

MatthewHerbst:
@OmgBrown sure, sorry for the trouble - happy to help with the rebase next time

Issue #95, opened 2019-02-07
Body cells appear incorrectly/blank when trying to print ag-grid


OmkarRa:
I use ag-grid library to show a table and react-to-print to print visible part of that table.

Print works correctly when horizontal scroll position is at the start.

But if I scroll it horizontally further, and then try to take print, only header cells appear correctly. Body cells are either blank or incorrect.

Here is a codesandbox that replicates the issue https://codesandbox.io/s/5wj8449ojx

Scroll horizontally
![image](https://user-images.githubusercontent.com/40591244/52397431-91924f00-2adb-11e9-8316-adbcab32445f.png)

Blank cells appear
![image](https://user-images.githubusercontent.com/40591244/52397448-a1aa2e80-2adb-11e9-84c6-d6ba4394a772.png)

MatthewHerbst:
Thanks for the report. I use ag-grid as well. I'll try to look into this over the weekend, though there are a few other bugs that I'd like to try and address first.

sameeravhad119:
Hi @MatthewHerbst,
I am also facing same issue. Could you please look into it ?

mikhail-angelov:
I have similar issue, sometime printed document is blank
I'm sure the root cause of this issue, not all html, style, link nodes are loaded once this timer is expired
```
    setTimeout(() => {
      target.contentWindow.focus()
      target.contentWindow.print()
      this.removeWindow(target)

      if (onAfterPrint) {
        onAfterPrint()
      }
    }, 500)
```
it's better remove this timer and count all html nodes, not just `link`
it will make lib more stable
I can prepare PR for it

duyhau206:
Hi every body. I have same issue. I just use div element to render component and sometime printed document is bank (where data is rendered). 
Looking forward to help from everyone

sameeravhad119:
@mikhail-angelov Thank you so much for your reply. I did changes locally as you mentioned. But Still it is giving me blank document for print.

MatthewHerbst:
Hi everyone. Sorry for the extreme delay on solving this. I had gotten stuck on it and shelved it, but recently thought of something and realized it was the problem.

By default, agGrid virtualizes columns and rows that are not displayed. What that means is that when `react-to-print` gets an internal copy of what's on the screen, `agGrid` does not actually render the columns to the right because from `react-to-print`'s perspective, they are not in frame.

[...]

PR #101, opened 2019-02-26
Body cells appear incorrectly/blank when trying to print ag-grid #95

mikhail-angelov:
wait for all new html nodes are loaded, before print document

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -40,16 +40,13 @@ class ReactToPrint extends React.Component {
     if (onBeforePrint) {
       onBeforePrint();
     }
+    target.contentWindow.focus();
+    target.contentWindow.print();
+    this.removeWindow(target);
 
-    setTimeout(() => {
-      target.contentWindow.focus();
-      target.contentWindow.print();
-      this.removeWindow(target);
-
-      if (onAfterPrint) {
-        onAfterPrint();
-      }
-    }, 500);
+    if (onAfterPrint) {
+      onAfterPrint();
+    }
   }
 
   handlePrint = () => {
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -40,16 +40,13 @@ class ReactToPrint extends React.Component {
     if (onBeforePrint) {
       onBeforePrint();
     }
+    target.contentWindow.focus();
+    target.contentWindow.print();
+    this.removeWindow(target);
 
-    setTimeout(() => {
```
MatthewHerbst, line 44: Why did you remove the `setTimeout` here? I'm not sure exactly why we needed it before, but it's been in the codebase almost since the first commit. @gregnb can you give any context on this please?

MatthewHerbst (commented):
Thanks for the PR! I like what you've done here. I left one question to be addressed.

Also, now that DOM nodes will be going into `markLoaded` as well, the param there should probably be changed to just be `node`, and the error message should be updated to be more generic, since now a DOM node could fail to load, not just a link node.

src/index.js
gregnb: hey @MatthewHerbst honestly I believe it because some browsers had issues if you tried to say close the window you were printing or in this case remove the element you are printing too quickly. Throwing in a hacky delay was how I found a lot of people to solve this issue:

https://stackoverflow.com/questions/6460630/close-window-automatically-after-printing-dialog-closes/19115657

if we don't think it's necessary anymore I'm ok with that but I would be just a little cautious here

MatthewHerbst: Thanks! @mikhail-angelov sounds like we should keep the `setTimeout` for now to be safe. I think it's unrelated to your core change anyways.

mikhail-angelov: Hi @MatthewHerbst you are right we should keep it
but actually my PR is not good enough, looks like my suggestion for root cause of this issue (sometime printed page is blank) is not correct, and I manage to reproduce the issue even with my fix, so I have to switch to print.js (it's more stable)
so, I close this PR, 
and probably make sense to investigate print.js and find difference between those 2 libs and make this lib more stable

mikhail-angelov:
unfortunately, it does not solve the issue :(

MatthewHerbst:
@mikhail-angelov sorry to hear that. If you have other thoughts on how to improve the library they would be greatly appreciated - a lot of us are trying to solve similar issues. Thanks!

PR #105, opened 2019-03-04
Update IE11 support

treago:
NodeList object spreading is not supported in IE11.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -121,7 +121,7 @@ class ReactToPrint extends React.Component {
       }
 
       const canvasEls = domDoc.querySelectorAll('canvas');
-      [...canvasEls].forEach((node, index) => {
+      Array.prototype.forEach.call(canvasEls, (node, index) => {
         node.getContext('2d').drawImage(srcCanvasEls[index], 0, 0);
       });
 


MatthewHerbst:
Hi @treago, thanks for the PR.

While it is true that object rest spread is not supported in IE11, we use Babel to remove the spread in the built version of the library. If you do a clean install and look at `node_modules/react-to-print/lib/index.js` you will see that there is no object rest spread in that file.

What issue are you having that is making you think this is the problem?

treago:
Hi @MatthewHerbst 

While trying to use your plugin in IE11 I get an error "invalid attempt to spread non-iterable instance".
The error is provided with a built version of the library
The same page does not show any error while being opened in Chrome / FF / Edge

MatthewHerbst:
@treago Sorry that you're running into that. Would you mind please showing how you're using the library? Thanks!

sergeyshmakov:
Having the same issue about IE11 error - it would be wonderful to get release this PR. Thanks @treago for your work here.

MatthewHerbst:
@sergeyshmakov can you please check if the `...` exists in the installed version of the library? The file to check is `yourPackage/node_modules/react-to-print/lib/index.js`. As far as I can tell, we do properly remove the `...` when building the library, and so that is not actually the problem you are having.

sergeyshmakov:
@MatthewHerbst Spread operator transpiled to `n = _toConsumableArray(s.querySelectorAll("canvas"));`

Where `_toConsumableArray` is:
`function _toConsumableArray(e) {
    return _arrayWithoutHoles(e) || _iterableToArray(e) || _nonIterableSpread()
}`

And finally in this function `_arrayWithoutHoles` and `_iterableToArray` return `false` and `_nonIterableSpread` generate `throw new TypeError("Invalid attempt to spread non-iterable instance")`.

Here is transpiled functions that return `false` for argument of `NodeList<Canvas>` object.

`function _iterableToArray(e) {
    if (Symbol.iterator in Object(e) || "[object Arguments]" === Object.prototype.toString.call(e)) return Array.from(e)
}`

`function _arrayWithoutHoles(e) {
    if (Array.isArray(e)) {
        for (var t = 0, r = new Array(e.length); t < e.length; t++) r[t] = e[t];
        return r
    }
}`

sergeyshmakov:
@MatthewHerbst I fixed this issue using `Array.from()` insted of `...` spread operator in my working copy (tsconfig `"target": "es5"` and `"lib": [ "es2015", "es2016", "es2017", "dom" ]`)

MatthewHerbst:
@sergeyshmakov [`Array.from`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from) is definitely not supported by IE11...

MatthewHerbst:
Ok, so I think the real issue here is that `NodeList` is Array-like, but is not itself an array, and so in some older browsers is missing `Symbol.iterator`. I'm surprised Babel is doing this considering we specify IE >= 10.

MatthewHerbst:
The `Symbol.iterator` check still exists in the code after applying @treago's changes. @treago were you able to verify that your change works on IE, using the _built_ version of the code in `lib/index.js`?

sergeyshmakov:
> @sergeyshmakov [`Array.from`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from) is definitely not supported by IE11...

Because of this I wrote to you ts-config settings.

sergeyshmakov:
@MatthewHerbst Do you really thing that in this place is necessary to create a new Array object?

I think this is ok to use NodeList collection directly as it is. Because @treago code is just calling `forEach` directly from `NodeList<Canvas>`. And I checked it also - this code is works.

treago:
> Ok, so I think the real issue here is that `NodeList` is Array-like, but is not itself an array, and so in some older browsers is missing `Symbol.iterator`. I'm surprised Babel is doing this considering we specify IE >= 10.

@MatthewHerbst sry for keeping you waiting for a response. Yes, after expanded investigation I figured out the same issue. Seems like Babel does not handle `Symbol.iterator` for IE11

For now, my PR is not valid but an issue is still actual.

gregnb:
Issue: https://github.com/babel/babel/issues/7597

PR: https://github.com/babel/babel/pull/8947

MatthewHerbst:
@gregnb thanks! This issue must have started then after I upgraded us to Babel 7 :( I posted in the PR to see what the status is

phatlt:
Sorry, but did we have any solution?

MatthewHerbst:
@phatlt The issue is an upstream issue in Babel. We have three options to deal with it:

1. use an older version of this library
2. we can rollback the Babel 7 change (could result in other issues)
3. we just wait

I suggest you post in the linked thread to encourage the Babel team to accept the PR that would fix this issue.

sergeyshmakov:
@MatthewHerbst Why you cant just replace `...` to `Array.from`?

sergeyshmakov:
I wrote to you about Typescript because `Array.from` nicely transpiled there.
It takes 5 min to rewrite your small app to Typescript and set `"target": "es5"`  with libs `"lib": [ "es2015", "es2016", "es2017", "dom" ]` and dont use Babel anymore.

MatthewHerbst:
@sergeyshmakov converting this package to TypeScript is a decision @gregnb would have to make. My thoughts are that doing so to solve a single problem is serious overkill, and that it will simply introduce other problems now or in the future.

sergeyshmakov:
@MatthewHerbst Hmm, problems with Typescript? Babel have much more unstable platform for transpiling. On your thoughts it is better to live now with breaking changes than fix it small app without any problems. Dont see any dangerous in Typescript.

MatthewHerbst:
@sergeyshmakov you are welcome to make a PR changing the app to TypeScript and we can discuss it there.

MatthewHerbst:
Going to close this in favor of #111 which should solve this issue and possibly a few others related to IE as well

PR #114, opened 2019-04-02
add prettier support

gregnb:
@MatthewHerbst Somehow forgot to add in prettier when I created this project. Better late than never!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "react-to-print",
-  "version": "2.0.0-alpha.7",
+  "version": "2.1.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
@@ -6233,6 +6233,12 @@
       "version": "0.1.1",
       "resolved": "https://registry.npmjs.org/posix-character-classes/-/posix-character-classes-0.1.1.tgz",
       "integrity": "sha1-AerA/jta9xoqbAL+q7jB/vfgDqs=",
+      "dev": true
+    },
+    "prettier": {
+      "version": "1.16.4",
+      "resolved": "https://registry.npmjs.org/prettier/-/prettier-1.16.4.tgz",
+      "integrity": "sha512-ZzWuos7TI5CKUeQAtFd6Zhm2s6EpAD/ZLApIhsF9pRvRtM1RFo61dM/4MSRUA0SuLugA/zgrZD8m0BaY46Og7g==",
       "dev": true
     },
     "pretty-error": {
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "html-webpack-plugin": "^3.2.0",
     "husky": "^1.3.1",
     "lint-staged": "^7.3.0",
+    "prettier": "^1.16.4",
     "react": "^16.8.4",
     "react-dom": "^16.8.4",
     "ts-loader": "^5.3.3",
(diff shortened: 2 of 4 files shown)


MatthewHerbst (approved):
Nice! Are all the lint rules passing with this? We'll have to make sure there is no conflict between Prettier and tslint.

Also, while you're at it, can you remove the `.eslintignore` file please (assuming it's not used by tslint). We forgot to remove it with the TS commit

gregnb:
Is tslint working for you? I'm looking at the master branch as it is always complaining about `tslint.json` not being found. I have zero typescript experience. Is this something we are missing or something that gets generated in a step I'm not aware of?

```


Invalid option for configuration: tslint.json
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! react-to-print@2.1.0 lint: `tslint -c tslint.json 'src/**/*.{ts,tsx}'`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the react-to-print@2.1.0 lint script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
```

MatthewHerbst:
Hmm, looks like we forgot to add the `tslint.json` file to the repo when the lint step was added (I have a global file locally so it was falling back to that so I didn't notice). @cmckni3 / @sergeyshmakov do either of you have a file that we can use for this lying around?

gregnb:
@MatthewHerbst any issue with using your copy? if not let's get it in this PR

cmckni3:
Well tslint is merging with eslint soon. Here's an example that I commonly use in projects:

```
{
  "extends": [
    "tslint:recommended",
    "tslint-config-prettier"
  ],
  "rules": {
    "array-type": false,
    "arrow-return-shorthand": true,
    "callable-types": true,
    "class-name": true,
    "deprecation": {
      "severity": "warn"
    },
    "forin": true,
    "interface-name": false,
    "interface-over-type-literal": true,
    "label-position": true,
    "max-classes-per-file": false,
    "member-access": [
      true,
      "no-public"
    ],
    "member-ordering": [
      true,
      {
        "order": [
          "static-field",
          "instance-field",
          "static-method",
          "instance-method"
        ]
      }
    ],
    "no-arg": true,
    "no-bitwise": true,
    "no-console": [
      true,
      "debug",
      "info",
      "time",
      "timeEnd",
      "trace"
    ],
    "no-construct": true,
    "no-debugger": true,
    "no-duplicate-super": true,
    "no-empty": false,
    "no-empty-interface": true,
    "no-eval": true,
    "no-inferrable-types": [
      true,
      "ignore-params"
    ],
    "no-misused-new": true,
    "no-non-null-assertion": true,
    "no-shadowed-variable": true,
    "no-string-literal": false,
    "no-string-throw": true,
    "no-switch-case-fall-through": true,
    "no-unnecessary-initializer": true,
    "no-unused-expression": true,
    "no-unused-variable": true,
    "no-use-before-declare": true,
    "no-var-keyword": true,
    "object-literal-sort-keys": false,
    "prefer-const": true,
    "radix": true,
    "triple-equals": [
      true,
      "allow-null-check"
    ],
    "unified-signatures": true,
    "variable-name": false
  }
}
```

cmckni3:
Some of those rules don't apply for TS >= 3.0.0 (examples: `no-unused-expression` and `no-unused-variables`).

tslint has been flip flopping on them though.

Issue #115, opened 2019-04-04
v.2.1.0 Uncaught Invariant Violation: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: object.


karpovich-philip:
I can't render my component which contains ReactToPrint. I got an error which I mentioned in title. It tells me "Check your code at {MyComponent}.jsx:{rowNumber}. Row Number it's exactly ReactToPrint component.

But it works fine with previous version 2.0.0-alpha.7

Dependencies:
"react": "^16.5.2",
"react-redux": "^5.0.6",
"react-router": "^4.3.1",
"react-router-dom": "^4.3.1",
"react-dom": "^16.5.2",

duyhau206:
i use:
"prop-types": "^15.5.10"
"react": "^16.8.3",
"react-dom": "^16.2.0",
"react-to-print": "^2.0.0-alpha.7

And I have same warning

MatthewHerbst:
Hi @karpovich-philip, @duyhau206. Can you please show how you import the component?

@sergeyshmakov is there anything from the TypeScript PR that might be causing this?

karpovich-philip:
@MatthewHerbst 
`import ReactToPrint from 'react-to-print';`

duyhau206:
I import component as below
`import ReactToPrint from 'react-to-print';`

mikdatdogru:
i have same error too. `2.0.0-alpha-6` version is working



[codesandbox](https://codesandbox.io/s/kmmw7l39y7) is still working because package.json updated, but still `2.0.0-alpha-1` version included

![image](https://user-images.githubusercontent.com/10121255/55553356-a0b30900-56e8-11e9-9e63-93eaec41e765.png)

dsuhic:
I'm guessing something's wrong with the exports. Managed to get it working by using the component as `<ReactToGrid.default>`. Strangely, simply changing the import statement to `import { default as ReactToPrint } from 'react-to-print'` doesn't work.

MatthewHerbst:
I'm able to run the example app in this lib by installing `react-to-print` locally and importing that in the example instead of importing the source file directly.

Clearly an export issue. Does anyone see an issue with the code?

duyhau206:
Sometime body print page was bank when i trying print with version 2.0.0-alpha-4. 
About version 2.1.1 I have not imported it yet so I haven't checked it yet

alihameedaukla:
[...]

PR #116, opened 2019-04-07
Fixed import bug

aviklai:
Fix for #115.

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
   "scripts": {
     "start": "webpack-dev-server --env.development --progress --colors --hot --open",
     "build": "webpack --env.production --progress",
+    "prepare": "npm run build",
     "lint": "tslint -c tslint.json 'src/**/*.{ts,tsx}'"
   },
   "repository": {
--- a/webpack.prod.config.js
+++ b/webpack.prod.config.js
@@ -9,7 +9,10 @@ module.exports = {
     output: {
         filename: 'index.js',
         path: path.resolve(__dirname, './lib'),
-        libraryTarget: 'commonjs'
+        libraryTarget: 'umd',
+        library: 'lib',
+        umdNamedDefine: true,
+        globalObject: `(typeof self !== 'undefined' ? self : this)`
     },
     externals : {
         'react': 'react',


MatthewHerbst:
Thanks for the PR! Can you explain why this fixes it please? Changes make sense, just not sure what the issue with the current setup is

aviklai:
@MatthewHerbst Hi, unfortunately I am not sure either why the previous setup didn't work but it looked like an issue with the entry point. I stumbled upon the issue after having the exact same error on my project.

I checked this new setup on a project I am working on and it looks o.k.
You can try to download from my fork and check that it works.

hendra-go:
I can confirmed that this PR works.
But i think like @MatthewHerbst said, this is an export issue, not import issue, isn't it?

And i think is because UMD (Universal Module Definition) is well... more  "universal" and compatible with both AMD and CommonJS

MatthewHerbst:
> And i think is because UMD (Universal Module Definition) is well... more "universal" and compatible with both AMD and CommonJS

Was the previous build (before 2.1) UMD? @sergeyshmakov did the TS PR change the build type?

MatthewHerbst:
@gregnb do you have any knowledge here? I'm thinking we merge this unless anyone thinks it will cause other problems. @cmckni3 do you see any issues with the existing TS config, or with this PR?

cmckni3:
Ah yes. The TS conversion added `libraryTarget: "commonjs"` to the webpack. UMD should be published to npm so it can be consumed in multiple environments (commonjs, AMD, ESM, etc).

MatthewHerbst:
@gregnb can you please publish this fix as `2.1.1`? thanks!

Thanks @cmckni3!

gregnb:
Soon as I land from vacation Saturday. Sorry for the delay

gregnb:
@MatthewHerbst I'm still swamped from traveling. I've added you to have publishing access on npm. Can you test let me know if you have any issues?

MatthewHerbst:
v2.1.1 is published. Please report any problems. Thank you everyone for the help fixing this!

(Thanks for the privileges @gregnb!)

gregnb:
Awesome! No problem thanks for all your help :)

Issue #118, opened 2019-04-12
Image not getting displayed on Print preview (Mac)


SaravananAnbu:
Once I click print to print a file from web, then the print preview dialog box opens, but images on my file is missing in the print file, hence after exporting as pdf, images are missing from the print file. Also every time I click print, the results vary as some times it picks up one or two images or it picksup nothing or it picksup everything.

MatthewHerbst:
Do you have the "Background graphics" option turned on in your print dialog?

SaravananAnbu:
Yes @MatthewHerbst

MatthewHerbst:
Can you please make a codesandbox or something replicating the issue?

SaravananAnbu:
`import React, {Fragment} from 'react';
import ReactToPrint from 'react-to-print';
import _ from 'lodash';
import Config from '../../config';
import history from '../../history';

const longTextArray = [ "Custom Notes", "others" ]

class PrintSinglePage extends React.Component {
    constructor(props) {
        super(props);
        this.state = {
            Name: '',
            Description: '',
            Notes: '',
            Quality_Notes: '',
            styleData:[],
            detailCards: [],
            styleColors: [],
            Original_Samp_Size: '',
            sizeEquivalenceCountry:'',
            productionSampCountry: '',
            Size_equivalence: [],
            Production_samp_size: [],
            loading: true,
            styleData: {},
            qualityLen: 0,
            notesLen: 0,
            count: 0
        }
    }
    componentDidUpdate(prevProps) {
        const styleData = this.props.styleData;
        if (this.props.styleData !== prevProps.styleData) {
            if (parseInt(this.props.styleId) === parseInt(this.props.styleData.Id)) {
                const notesLen = styleData.Notes.split(/\r\n|\r|\n/)
                const qualityLen = styleData.Quality_Notes.split(/\r\n|\r|\n/)
                this.setState({
                    Name: styleData.Name,
                    Description: styleData.Description,
                    Notes: styleData.Notes,
                    Quality_Notes: styleData.Quality_Notes,
                    Original_Samp_Size: styleData.Original_Samp_Size,
                    Production_samp_size: styleData.Production_samp_size,
                    Product_size: styleData.Product_size,
                    Size_equivalence: styleData.Size_equivalence,
                    sizeEquivalenceCountry: styleData.sizeEquivalenceCountry,
                    productionSampCountry: styleData.productionSampCountry,
[...]

PR #124, opened 2019-05-03
Fix for loading fonts

aviklai:
Fix for #118.
Before the rewrite to typescript, the triggerPrint settimout had a 500 milliseconds delay, which gave the browser enough time to load the fonts. 
In the new version, the delay was 0 milliseconds, and that's why the fonts weren't  loaded.


MatthewHerbst:
Thanks for finding that! Been scratching my head about it since the first report

MatthewHerbst:
Just published v2.1.2 with this fix, thanks again!

Issue #64, opened 2018-10-31
Print dialogue only open once in safari on Mac


kapssoni:
Please help with following issue.

On safari, first time print dialogue opens up and everything works fine.
When trying to print again, A dialogue coming up asking to print or not(That is safari may be)
But after i click on "Print" then print dialogue is not showing up.

I also tried the same thing on the example provided in repo. same thing.

Please guide if you faced such issue or help fixing it.

MatthewHerbst:
I was able to print twice using the demo app in Safari v12, MacOS 10.14 (Mojave).

Can you please me more specific about how to reproduce your issue, or even upload a video if possible?

Also, what version of Safari are you on?

MatthewHerbst:
@kapsonic any updates on this by chance please?

kapssoni:
To make this work I had to make few changes the way iframe is appended to the body and removed.

There us confirmation on safari to print and by the time user confirm to print the iframe is removed from dom. That is why this was happening. I changed and it worked.

MatthewHerbst:
@kapsonic Great that you got it working! Could you please show us what you changed so that we can help others with this issue in the future?

PR #129, opened 2019-05-22
Fix issue #64: Print dialogue only open once in safari on Mac

eleonora-guidi:
Removed removeWindow method. Tested on example and custom app, seems to work fine on Chrome, Firefox and Safari.

Updated version on package.json and Changelog file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## 2.1.3 (May 22nd, 2019)
+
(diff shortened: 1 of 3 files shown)


eleonora-guidi:
One user had already attempted this in the past but was never merged in master and the issue was still present. 
See the pull request #56

MatthewHerbst:
This is a duplicate of #56. It has several problems, and does not actually completely work by itself. Please feel free to discuss in the #56 thread.

PR #134, opened 2019-05-31
Fix issue #64: Print dialogue only open once in safari on Mac

eleonora-guidi:
Alternative solution to PR #56

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,5 @@
 # CHANGELOG
 
-## 2.1.3 (May 22nd, 2019)
-
-- FIX [64](https://github.com/gregnb/react-to-print/issues/64): Print dialogue only open once in safari on Mac
-
 ## 2.1.2 (May 3rd, 2019)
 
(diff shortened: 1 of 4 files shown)


MatthewHerbst (changes requested):
Thanks for the PR! I'll test it out this weekend. One thing: can you please remove the CHANGELOG and version changes? Those will be in their own PR

eleonora-guidi:
> Thanks for the PR! I'll test it out this weekend. One thing: can you please remove the CHANGELOG and version changes? Those will be in their own PR

Hi! Thank you, if you have time to test it in the weekend it would be really great! 
I have reverted the changes on the changelog and package.jsoon as requested. Hope it's fine.

MatthewHerbst (approved):
I've tested this on desktop Chrome and Safari and I've had others test it on mobile Safari on both iPad and iPhone, and it seems to work.

Apologies for the delay - I don't have an iOS device myself so it took a while to get it tested on mobile Safari.

Going to merge and publish, hopefully it solves issues for everyone!

MatthewHerbst:
v2.1.3 is published! Please let me know if you run into an other issues

Issue #140, opened 2019-06-28
UpdatedDom ignored on onbeforeprint


LeopoldLerch:
Hi, I am using this lines

```jsx
<ReactToPrint
  trigger={() => <a href="#">Print this out!</a>}
  content={() => componentRef.current}
  onBeforePrint={() => setPrintMode(true)}
  onAfterPrint={() => setPrintMode(false)}
/>
```

where i am setting Printmode to true, in order to render additional Content (Legend) for the print.

```jsx            
{printMode &&
  <div>
    {data.legende.map((l) => <div>{`${l.abk}: ${l.text}`}</div>)}
  </div>
}
```

However. It is called and the DOM updated (as i can see in the background, but in the print window it still shows the "non-print-version". 

the full thing I use:

```jsx
const componentRef = useRef();

return (
  <div className={css(styles.formsContainer)} ref={componentRef} >
    <h2 className={styles.formsTitle}>
      {Strings.PressespiegelHeaderText}
      {!printMode &&
        <ReactToPrint
          trigger={() => <a href="#">Print this out!</a>}
          content={() => componentRef.current}
          onBeforePrint={() => setPrintMode(true)}
          onAfterPrint={() => setPrintMode(false)}
        />
      }
      </h2>
      {data &&
        <div>
          {`Ausgabe vom ${data.datum}`}
          {data.kapitel.slice(1).map((kapitel) => {
            return (
              <div key={kapitel.id}>
                <div>
                  {kapitel.bezeichnung}
                </div>
                <ListBase
                  hideArrow
                  items={kapitel.KM}
                  renderItem={renderKM}
                />
              </div>
            );
          })}
          {printMode &&
            <div>
              {data.legende.map((l) => <div>{`${l.abk}: ${l.text}`}</div>)}
            </div>}
            {data.copyright}
        </div>
      }
    </div>
);`

MatthewHerbst:
Hello. Can you please show what `setPrintMode` does?

LeopoldLerch:
It just sets a state property to true so that the rendering can react and
render the additional elements out too

Matthew Herbst <notifications@github.com> schrieb am Fr., 28. Juni 2019,
20:05:

MatthewHerbst:
[...]

PR #146, opened 2019-07-19
Allowing onBeforePrint to return a Promise

aviklai:
#140

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -14,7 +14,7 @@ export interface IReactToPrintProps {
     /** Copy styles over into print window. default: true */
     copyStyles?: boolean;
     /** Callback function to trigger before print */
-    onBeforePrint?: () => void;
+    onBeforePrint?: () => void | Promise<any>;
     /** Callback function to trigger after print */
     onAfterPrint?: () => void;
     /** Override default print window styling */
@@ -29,13 +29,7 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
     linksLoaded: Element[];
     linksErrored: Element[];
 
-    triggerPrint = (target) => {
-        const { onBeforePrint, onAfterPrint } = this.props;
-
-        if (onBeforePrint) {
-            onBeforePrint();
-        }
-
+    startPrint = (target, onAfterPrint) => {
         setTimeout(() => {
             target.contentWindow.focus();
             target.contentWindow.print();
@@ -44,6 +38,21 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
                 onAfterPrint();
             }
         }, 500);
+    }
+
+    triggerPrint = (target) => {
+        const { onBeforePrint, onAfterPrint } = this.props;
+
+        if (onBeforePrint) {
+            const onBeforePrintOutput = onBeforePrint();
+            if (onBeforePrintOutput && typeof onBeforePrintOutput.then === 'function') {
+                onBeforePrintOutput.then(() => {
+                    this.startPrint(target, onAfterPrint);
+                });
+            } else {                
+                this.startPrint(target, onAfterPrint);
+            }
+        }            
     };
 
     handlePrint = () => {
--- a/README.md
+++ b/README.md
@@ -135,7 +135,7 @@ The component accepts the following props:
 |     **`trigger`**     | function | A function that returns a React Component or HTML element                                                                           |
 |     **`content`**     | function | A function that returns a component reference value. The content of this reference value is then used for print                     |
 |   **`copyStyles`**    | boolean  | Copies all &lt;style> and &lt;link type="stylesheet" /> from <head> inside the parent window into the print window. (default: true) |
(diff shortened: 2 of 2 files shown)


src/index.tsx
```diff
@@ -44,6 +38,23 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
...
+
+    triggerPrint = (target) => {
+        const { onBeforePrint, onAfterPrint } = this.props;
+
+        if (onBeforePrint) {
+            const onBeforePrintOutput = onBeforePrint();
+            if (onBeforePrintOutput instanceof Promise) {
+                onBeforePrintOutput.then(r => {
+                    this.startPrint(target, onAfterPrint);
+                }).catch(e => {
```
MatthewHerbst: I don't think we should `catch` this failure. If we do this then if the user chains a `.catch` it won't get fired.

src/index.tsx
```diff
@@ -44,6 +38,23 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
...
         }, 500);
+    }
+
+    triggerPrint = (target) => {
+        const { onBeforePrint, onAfterPrint } = this.props;
+
+        if (onBeforePrint) {
+            const onBeforePrintOutput = onBeforePrint();
+            if (onBeforePrintOutput instanceof Promise) {
+                onBeforePrintOutput.then(r => {
```
MatthewHerbst: Nit: `r` isn't used and so doesn't need to be defined, should just be `()`

src/index.tsx
```diff
@@ -14,7 +14,7 @@ export interface IReactToPrintProps {
     /** Copy styles over into print window. default: true */
     copyStyles?: boolean;
     /** Callback function to trigger before print */
-    onBeforePrint?: () => void;
+    onBeforePrint?: () => void | Promise<any>;
```
MatthewHerbst, line 17: Please also update the README

src/index.tsx
```diff
@@ -44,6 +38,23 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
...
+
+        if (onBeforePrint) {
+            const onBeforePrintOutput = onBeforePrint();
+            if (onBeforePrintOutput instanceof Promise) {
+                onBeforePrintOutput.then(r => {
+                    this.startPrint(target, onAfterPrint);
+                }).catch(e => {
+                    console.error(`onBeforePrint was not resolved. error: ${e.toString()}`);
+                });
+            }else{
```
MatthewHerbst: Nit

```suggestion
            } else {                
```

MatthewHerbst (changes requested):
Thanks for jumping on this! Just a few minor things and then I think we'll be good to go (we'll have to release this under a breaking change but I think that's ok)

src/index.tsx
aviklai: @MatthewHerbst Done. Please tell me if the explanation is good enough.

aviklai:
@MatthewHerbst Why is this a breaking change? we still allow the onBeforePrint function to return void.

src/index.tsx
aviklai: Done.

src/index.tsx
aviklai: Done.

src/index.tsx
aviklai: Done.

MatthewHerbst:
This line:

```js
onBeforePrintOutput instanceof Promise
```

will break if `Promise` does not exist. Specifically, if you run that on a browser (I tested on IE10 via Browserstack) without `Promise`, you will get the error:

```js
var x = {}
x instanceof Promise
// Error: "'Promise' is undefined"
```

So the implementation as it would still require users to polyfill `Promise` for environments it doesn't exist in, which is not currently a library requirement, which is why it would require a breaking change.

Maybe there is another way to check if something is a `Promise`?

aviklai:
Apparently, there is no standardized way:
https://stackoverflow.com/questions/27746304/how-do-i-tell-if-an-object-is-a-promise

This is the es5 answer that we can use (and graphql-js use):
https://github.com/graphql/graphql-js/blob/master/src/jsutils/isPromise.js

So it would look like this:
`if (onBeforePrintOutput && typeof onBeforePrintOutput.then === 'function')`

What do you think?

MatthewHerbst:
I'm ok with that. It's not 100% perfect (as some commenters in the accepted answer of that SO post point out) but I think it's good enough

aviklai:
@MatthewHerbst Done - changed the way to check if it's a promise.

MatthewHerbst:
Awesome! One last request: can you please squash the commits together?

aviklai:
@MatthewHerbst Done.

MatthewHerbst (approved):
Awesome, thanks! I'll get a new version with this published in a little bit

MatthewHerbst:
Published as v2.2.0. Thanks again!

Issue #151, opened 2019-07-22
undefine closeAfterPrint in property


ywCn1998:
Hi
I use version 2.2.1
and It doesn't have closeAfterPrint property to use

MatthewHerbst:
Thanks for the report! I'm not sure when or why this was removed from the code but not the documentation, or if it was meant to stay in the code and was removed by mistake. I'll be able to take a closer look at this next week.

MatthewHerbst:
This has been corrected by the introduction of a new prop `removeAfterPrint` which replaces the old `closeAfterPrint`. This is now available in version 2.3.0, just published. Thanks for finding this issue!

PR #152, opened 2019-07-27
Added removeAfterPrint

aviklai:
#151

I looked at the commits of the past year and I saw that the closeAfterPrint property was used when the printing was done using a window and not an iframe.
Since currently the library uses iframe for printing, I added a removeAfterPrint property which removes the iframe after printing.
For now, it defaults to false since by default the iframe was not removed and I don't want it to add a breaking change (maybe someone used the iframe after the print)

### DIFF
--- a/example/index.tsx
+++ b/example/index.tsx
@@ -22,6 +22,7 @@ class Example extends React.Component {
                     content={this.renderContent}
                     onBeforePrint={this.handleBeforePrint}
                     onAfterPrint={this.handleAfterPrint}
+                    removeAfterPrint
                 />
                 <ComponentToPrint ref={this.setRef}/>
             </div>
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -21,6 +21,8 @@ export interface IReactToPrintProps {
     pageStyle?: string;
     /** Optional class to pass to the print window body */
     bodyClass?: string;
+     /** Optional - remove the iframe after printing.*/
+    removeAfterPrint?: boolean
 }
 
 export default class ReactToPrint extends React.Component<IReactToPrintProps> {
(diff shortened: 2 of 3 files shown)


example/index.tsx
```diff
@@ -22,6 +22,7 @@ class Example extends React.Component {
                     content={this.renderContent}
                     onBeforePrint={this.handleBeforePrint}
                     onAfterPrint={this.handleAfterPrint}
+                    removeAfterPrint={true}
```
MatthewHerbst: Nit: no need to be explicit when setting a bool
```suggestion
                    removeAfterPrint
```

src/index.tsx
```diff
@@ -33,7 +35,9 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
         setTimeout(() => {
             target.contentWindow.focus();
             target.contentWindow.print();
-            
+            if (this.props.removeAfterPrint) {
```
MatthewHerbst: I think this should come after the `if` below, so that `onAfterPrint` can interact with the `iframe` if the user needs it to?

MatthewHerbst (changes requested):
Thanks again for picking up my slack! Vacation really makes me lazy :)

example/index.tsx
aviklai: Done.

src/index.tsx
aviklai: You are right - I moved it to be after onAfterPrint.

aviklai:
@MatthewHerbst And enjoy your vacation :)

MatthewHerbst (approved):

MatthewHerbst:
Published as 2.3.0, thanks!

PR #154, opened 2019-08-06
Revamp linting, fix bug found from linting

MatthewHerbst:
This fixes our linting such that `npm run lint` now works, and from that it fixes a bug that was found with linting working:

When we encountered a link that was not a STYLE tag and that had an `href` value, we would override the loop going through all the stylesheets with an inner loop, possibly causing us to miss some style sheets or to attempt to access stylesheets that did did not exist on the page.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -136,9 +136,9 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
             }
 
             const canvasEls = domDoc.querySelectorAll("canvas");
-            for (let index = 0, l = canvasEls.length; index < l; ++index) {
-                const node = canvasEls[index];
-                node.getContext("2d").drawImage(srcCanvasEls[index] as HTMLCanvasElement, 0, 0);
+            for (let i = 0, canvasElsLen = canvasEls.length; i < canvasElsLen; ++i) {
+                const node = canvasEls[i];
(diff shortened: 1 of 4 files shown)


src/index.tsx
```diff
@@ -139,44 +144,48 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
...
                         const newHeadEl = domDoc.createElement(node.tagName);
                         const sheet = (node as HTMLStyleElement).sheet as CSSStyleSheet;
 
                         if (sheet) {
                             let styleCSS = "";
-                            for (let i = 0; i < sheet.cssRules.length; i++) {
-                                if (typeof sheet.cssRules[i].cssText === "string") {
-                                    styleCSS += `${sheet.cssRules[i].cssText}\r\n`;
+                            // NOTE: for-of is not supported by IE
+                            for (let j = 0, cssLen = sheet.cssRules.length; j < cssLen; j++) {
```
davegw: The outer loop increments with `++i` while the inner loop increments by `j++` (also uses `++j` below). Can we please make this consistent? Could see this leading to a bug down the line.

davegw (approved):
LGTM 👍

MatthewHerbst:
I accidentally include another bug fix in this PR.

We were seeing this error: `TypeError: Object doesn't support property or method 'remove'`

The fix is to remove from the parent node, rather than trying to remove the node directly. We already use this method for removing the iframe if it already exists when trying to print a second time.

PR #158, opened 2019-08-08
Create method onBeforeGetContent

andfs:
Sometimes is useful to update the component to print before printing, for example, when someone needs to hide somethings in the printing. 
The component triggers the method `getContent` an then calls `onBeforePrinting`, but at this time, the component is mounted and everything that you do in `onBeforePrinting` is not reflected in the component. 
So with `onBeforeGetContent` is possible to update the component before printing


src/index.tsx
```diff
@@ -209,10 +211,14 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
     public render() {
         const {
             trigger,
+            onBeforeGetContent,
         } = this.props;
 
         return React.cloneElement(trigger(), {
-            onClick: this.handlePrint,
+            onClick: () => onBeforeGetContent ? onBeforeGetContent().then((e) => {
```
MatthewHerbst: This will require a check similar to [here](https://github.com/gregnb/react-to-print/blob/67df094343115d35f002ef66a752abb3a6b9d261/src/index.tsx#L57) so that we can handle people who do not have browsers that support Promise.

src/index.tsx
```diff
@@ -209,10 +211,14 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
...
+                e();
```
MatthewHerbst: What is `e` here? How do you know that the user's implementation will return anything at all?

MatthewHerbst (changes requested):
Hi, thanks for the PR. A couple questions about the code. I like the idea, though we need to flesh it out a bit with how it would work.

Can you please update the docs as part of the PR, and, please provide an example of how this would work?

MatthewHerbst:
Also, we do not use Yarn for this package, so please remove the `yarn.lock` file

andfs:
> Hi, thanks for the PR. A couple questions about the code. I like the idea, though we need to flesh it out a bit with how it would work.
> 
> Can you please update the docs as part of the PR, and, please provide an example of how this would work?

Hi @MatthewHerbst . I'll do that :)

src/index.tsx
andfs: you're right

I'll improve to this.
Thank you :)

src/index.tsx
andfs: e is the function returned by the promise.
Do you think that changing the type of `onBeforeGetContet` to
`onBeforeGetContent?: () => Promise<() => void>;` 
would be enought? That way if we had a value in _onBeforeGetContent_ we'll know that hav to be a function or the user will get a compiler issue.

MatthewHerbst: But why do we need `e` at all? The user does whatever they need to do in `onBeforeGetContent` and then we continue, we don't need to do anything with a return value

MatthewHerbst:
Please don't change the formatting of the entire file. Besides the fact that we have lint rules which will break, it makes the diff impossible to read

src/index.tsx
andfs: Well, In my case I needed to update the state of a component to hide some information and this would need to be a promise to ensure the update in the component. So what I did in my component was:

```js
onBeforeGetContent = () => {
  return Promise.resolve(() => this.setState({ texto: "something" }));
}
```

how would be the correct way to do this?

MatthewHerbst: `setState` does not return anything, and `Promise.resolve` returns a Promise, not a function. A bigger issue with the above is that `react-to-print` might continue with getting the content before the `setState` has taken effect since `setState` is async.

I would change it to:

```js
onBeforeGetContent = () => {
  return new Promise((resolve, reject) => {
    this.setState({ texto: "something" }, resolve);
  });
}
```

That way you ensure that `setState` has been committed before the Promise resolves.

andfs: I think that everything is ok in my last PR

README.md
```diff
@@ -140,6 +140,7 @@ The component accepts the following props:
 |     **`content`**     | function | A function that returns a component reference value. The content of this reference value is then used for print                     |
 |   **`copyStyles`**    | boolean  | Copies all &lt;style> and &lt;link type="stylesheet" /> from <head> inside the parent window into the print window. (default: true) |
 |  **`onBeforePrint`**  | function | Optional callback function that triggers before print. Either returns void or a Promise. If the function returns a Promise the content will be printed when the Promise is resolved. Users are responsible for catching the Promise rejecting.                                                                                     |
+| **`onBeforeGetContent`** | function | Optional callback function that triggers before lib get content to print. Either returns a Promise with a function. The content will be printed after the function returns. It's useful for manipulate the component to print before printing.
```
MatthewHerbst: Should be:

Optional callback function that triggers before the library gathers the page's content. Either returns void or a Promise. Users are responsible for catching the Promise rejecting. This can be used to change the content on the page before printing.

README.md
```diff
@@ -140,6 +140,7 @@ The component accepts the following props:
 |     **`content`**     | function | A function that returns a component reference value. The content of this reference value is then used for print                     |
 |   **`copyStyles`**    | boolean  | Copies all &lt;style> and &lt;link type="stylesheet" /> from <head> inside the parent window into the print window. (default: true) |
 |  **`onBeforePrint`**  | function | Optional callback function that triggers before print. Either returns void or a Promise. If the function returns a Promise the content will be printed when the Promise is resolved. Users are responsible for catching the Promise rejecting.                                                                                     |
```
MatthewHerbst: Let's update this to:

Optional callback function that triggers before print. Either returns void or a Promise. If the function returns a Promise the content will be printed when the Promise is resolved. Users are responsible for catching the Promise rejecting. Note: this function is run immediately prior to printing, but after the page's content has been gathered. To modify content before printing, use `onBeforeGetContent` instead.

example/ComponentToPrint/index.tsx
```diff
@@ -11,7 +14,7 @@ export default class ComponentToPrint extends React.Component {
     }
 
     setRef = ref => this.canvasEl = ref;
-
+
```
MatthewHerbst: There should be no extra whitespace here

example/index.tsx
```diff
@@ -4,13 +4,31 @@ import * as ReactDOM from 'react-dom';
...
+            text: "000000000";
+        };
+    }
+
     handleAfterPrint = () => console.log('after print!');
     handleBeforePrint = () => console.log('before print!');
     renderContent = () => this.componentRef;
     renderTrigger = () => <button type="button">Print this out!</button>;
+    onBeforeGetContent = () => new Promise(resolve => {
+        this.setState({text: "text changed"}, () => resolve());
```
MatthewHerbst: Nit: `() => resolve()` can be changed to just `resolve`

MatthewHerbst: Please also handle the error case of the Promise rejecting (not that it would here, but the example should show users the full use-case)

example/index.tsx
```diff
@@ -22,9 +40,10 @@ class Example extends React.Component {
                     content={this.renderContent}
                     onBeforePrint={this.handleBeforePrint}
                     onAfterPrint={this.handleAfterPrint}
+                    onBeforeGetContent={this.onBeforeGetContent}
```
MatthewHerbst: Nit: this should go before `onBeforePrint`

src/index.tsx
```diff
@@ -209,10 +211,20 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
     public render() {
         const {
             trigger,
+            onBeforeGetContent
         } = this.props;
 
         return React.cloneElement(trigger(), {
-            onClick: this.handlePrint,
+            onClick: () => {
+                const onBeforeGetContentPromise = onBeforeGetContent();
```
MatthewHerbst: Nit: please change the name from `onBeforeGetContentPromise` to `onBeforeGetContentOutput` since we don't actually know if it is a Promise yet

MatthewHerbst: Also, you need to check that `onBeforeGetContent` is passed before trying to call it. This should be similar to https://github.com/gregnb/react-to-print/blob/67df094343115d35f002ef66a752abb3a6b9d261/src/index.tsx#L55

src/index.tsx
```diff
@@ -209,10 +211,20 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
...
+                if(onBeforeGetContentPromise && typeof onBeforeGetContentPromise.then === "function") {
+                    onBeforeGetContentPromise.then(() => {
+                    return this.handlePrint();
```
MatthewHerbst: The `return` here isn't needed - `this.handlePrint()` does not return anything

src/index.tsx
```diff
@@ -209,10 +211,20 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
...
+                    })
+                } else {
+                    return this.handlePrint();
```
MatthewHerbst: The `return` here isn't needed - `this.handlePrint()` does not return anything

MatthewHerbst (changes requested):

.vscode/launch.json
```diff
@@ -0,0 +1,15 @@
+{
```
MatthewHerbst: Please remove this file.

MatthewHerbst (changes requested):

README.md
andfs: ok I'´ll do it now.

src/index.tsx
andfs: I need to return the `this.handlePrint()` since all function is inside a **{}**
if nothing is returned, the function won't be called by React on onClick

src/index.tsx
andfs: I need to return the this.handlePrint() since all function is inside a {}
if nothing is returned, the function won't be called by React on onClick

src/index.tsx
```diff
@@ -17,6 +17,8 @@ export interface IReactToPrintProps {
     onBeforePrint?: () => void | Promise<any>;
     /** Callback function to trigger after print */
     onAfterPrint?: () => void;
+    /** Callback function to trigger before get content to print */
+    onBeforeGetContent?: () => Promise<any>;
```
MatthewHerbst: Should be: `() => void | Promise<any>;`

MatthewHerbst (changes requested):

src/index.tsx
MatthewHerbst: Ah right, but I see you now have changed it to a function, thank you

src/index.tsx
andfs: changed

README.md
andfs: changed

example/index.tsx
andfs: changed

.vscode/launch.json
andfs: changed

src/index.tsx
```diff
@@ -209,10 +211,24 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
...
+            onBeforeGetContent
         } = this.props;
 
         return React.cloneElement(trigger(), {
-            onClick: this.handlePrint,
+            onClick: () => {
+                if(onBeforeGetContent) {
+                    const onBeforeGetContentOutput = onBeforeGetContent();
+                    if(onBeforeGetContentOutput && typeof onBeforeGetContentOutput.then === "function") {
+                        onBeforeGetContentOutput.then(() => {
```
MatthewHerbst: Nit: can be simplified to: `onBeforeGetContentOutput.then(this.handlePrint)`

src/index.tsx
```diff
@@ -209,10 +211,24 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
...
+                        this.handlePrint();
+                        })
+                    } else {
+                        return this.handlePrint();
```
MatthewHerbst: Nit: this return is no longer needed

src/index.tsx
```diff
@@ -209,10 +211,24 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
...
+                    }
+                } else {
+                     return this.handlePrint();
```
MatthewHerbst: Nit: this return is no longer needed

example/index.tsx
```diff
@@ -4,13 +4,33 @@ import * as ReactDOM from 'react-dom';
...
+        };
+    }
+
     handleAfterPrint = () => console.log('after print!');
     handleBeforePrint = () => console.log('before print!');
     renderContent = () => this.componentRef;
     renderTrigger = () => <button type="button">Print this out!</button>;
+    onBeforeGetContent = () => new Promise((resolve, reject) => {
+        this.setState({text: "text changed"}, () => resolve);
+        //if something goes wrong, regect the promise:
```
MatthewHerbst: Nit: actually, feel free to remove this comment: I'm going to make a commit after merging this that makes it easier for users to handle the rejection case in both `onBeforePrint` and in `onBeforeGetContent`

MatthewHerbst (commented):
We're really close here, thanks for making all the changes, just a few small ones left.

example/index.tsx
andfs: changed

src/index.tsx
andfs: changed

src/index.tsx
andfs: I think we need this return because if the user's browser doesn't support promises, we will return the handlePrint method to onClick. Remember that we need to return since our function is inside **{}**

src/index.tsx
andfs: I think we need this return because if the user doesn't pass onBeforeGetContentOutput method, we will return the handlePrint method to onClick. Remember that we need to return since our function is inside **{}**

src/index.tsx
MatthewHerbst: If they don't support Promises then we just call `this.handlePrint();` which is fine. The arrow function you have is what `onClick` will call, not what is returned from that. `onClick` does not have a return value and so methods we call after `onClick` is called do not need one either.

src/index.tsx
MatthewHerbst: See my comment above

MatthewHerbst (changes requested):
Can you undo the formatting changes of the whole file please? Looking forward to getting this merged this week :)

src/index.tsx
andfs: changed

src/index.tsx
andfs: changed

MatthewHerbst (approved):
Looks great! Can you please just squash your commits now into a single commit and then I'll merge?

MatthewHerbst (approved):
🎉 awesome, thank you! Will merge now. I'm going to make another commit cleaning up a few things across the library. Hope to have a release with this included in it tonight or tomorrow!

MatthewHerbst:
Oh wait, one thing, please squash the commits first! Do you know how to do that?

andfs:
> Oh wait, one thing, please squash the commits first! Do you know how to do that?

I think I squashed in my last push, didn't I?

MatthewHerbst:
> > Oh wait, one thing, please squash the commits first! Do you know how to do that?
> 
> I think I squashed in my last push, didn't I?

Nope - there should only be a single commit after you squash, right now there are 13 commits.

andfs:
> > > Oh wait, one thing, please squash the commits first! Do you know how to do that?
> > 
> > 
> > I think I squashed in my last push, didn't I?
> 
> Nope - there should only be a single commit after you squash, right now there are 13 commits.

Sorry. I think I did it right.

I thought that was that: 
https://github.com/gregnb/react-to-print/pull/158/commits/74d61dfdb8e7abf817ccc78fb19213758124bda5

Could you tell me how to do it?

MatthewHerbst:
Let me know if you have any trouble from this: https://github.com/wprig/wprig/wiki/How-to-squash-commits

andfs:
> Let me know if you have any trouble from this: https://github.com/wprig/wprig/wiki/How-to-squash-commits

sorry, but I wasn't able to make it work :/

andfs:
> Let me know if you have any trouble from this: https://github.com/wprig/wprig/wiki/How-to-squash-commits

There is another way to squash it?

MatthewHerbst:
Can you add me to have permissions to push to your fork of `react-to-print`? I can fix the branch for you

andfs:
> Can you add me to have permissions to push to your fork of `react-to-print`? I can fix the branch for you

sure

MatthewHerbst:
I'm trying to push it up to your branch but it's not letting me. I can keep trying, or I can push it as a new PR and change the author on the commit to you. Would that be ok with you?

andfs:
> I'm trying to push it up to your branch but it's not letting me. I can keep trying, or I can push it as a new PR and change the author on the commit to you. Would that be ok with you?

of course, no problem :)

MatthewHerbst:
Squashed and ready to go in #160

MatthewHerbst:
Published as v2.4.0. Thanks so much for your work on this!

PR #160, opened 2019-08-20
Add new method `onBeforeGetContent`

MatthewHerbst:
This method allows the user to change the contents of the page before it is retrieved for printing. This occurs before the existing `onBeforePrint` which is called after the page content has been retrieved and right before the print window is opened.

Long term, it likely makes sense to change `onBeforePrint` to have this functionality, and then remove this function, since there is little use for it with this function existing.


MatthewHerbst (commented):
Excited to see this in action @andfs

andfs:
> Excited to see this in action @andfs

Thanks, man \o/

mckelveygreg:
Thank you all for you work :) I've bee watching this for the last week 🚀

MatthewHerbst:
Will get this released once I get #161 merged since right now a Promise rejection could result in the library still printing depending on if the user handled the error or not.

MatthewHerbst:
Published as v2.4.0!

andfs:
> Published as v2.4.0!

Great news \o/
Thanks :)

PR #161, opened 2019-08-20
Better promise rejection handling

MatthewHerbst:
Adds a new `onPrintError` callback method that can be used to respond to Promise rejection events.

Also includes some docs cleanup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,21 +130,22 @@ const Example = () => {
 
 ## API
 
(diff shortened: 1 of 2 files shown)


MatthewHerbst:
@andfs this is the follow-up PR I had mentioned. It adds a way for users to be able to get notified if the Promise rejects. Do you mind reviewing it for me?

Issue #127, opened 2019-05-14
How to print child component without render it on parent component


jogoool:
My parent component is a table and it has many other components with a print button that generate report to print, the child component <ReportTemplate /> is a template of report. when I try to reference it` <ReportTemplate ref={el => (this.componentRef = el)}/>` it is rendered in Parent component. How can I pass the child component to `<ReactToPrint/>` without rendering it on parent component.

example 

```
import ReportTemplate from './ReportTemplate';

class Report extends Component {
...

render() {
     return(
        <div>... </div> . <----- other child components
        <div>
            <ReactToPrint
                  trigger={() => <a href="#">Print this out!</a>}
                  content={() => this.componentRef}
            />
            <ReportTemplate ref={el => (this.componentRef = el)} /> . <--- ReportTemplate is rendering in parent component.
      </div>
    );
}
}

export default Report;
```

lxx7061:
```js
<ReportTemplate ref={el => (this.componentRef = el)} /> 
```
 to => 
```js
<div style={{display:none}}>
     <ReportTemplate ref={el => (this.componentRef = el)} /> 
</div>
```

fershibli:
> ```js
> <ReportTemplate ref={el => (this.componentRef = el)} /> 
> ```
> 
> to =>
> 
> ```js
> <div style={{display:none}}>
>      <ReportTemplate ref={el => (this.componentRef = el)} /> 
> </div>
> ```

I've tried your solution, but got an error instead:
`'none' is not defined  no-undef`

here's my code:
```
<div>
  <ReactToPrint
    trigger={() => <Button >{t("Print Token")}</Button>}
    content={() => this.componentToPrint}
  />
  <div style={{display:none}}>
    <ResultAccessToken ref={el => (this.componentToPrint = el)} />
  </div>
</div>
```

MatthewHerbst:
@FernandoLuizNemeChibli the `style` prop does not understand CSS, so values need to be strings:

`style={{display:none}}` -> `style={{ display: 'none' }}`

fershibli:
@MatthewHerbst oh! now worked perfectly! thanks! (:
Thank you too, @lxx7061!

phongle2512:
Actually, when I have more than 200 tables to print (each table per page), the `<ReportTemplate ref={el => (this.componentRef = el)} />` make the parent component render too slow :( although I wrap a div to hide it `<div style={{display:"none"}}></div>`

[...]

PR #167, opened 2019-09-18
Add FAQ for hiding ComponentToPrint

devuxer:
See https://github.com/gregnb/react-to-print/issues/127#issuecomment-532465420


MatthewHerbst (approved):
Thanks!

PR #178, opened 2019-11-05
Added getPrintHandler-callback to programmatically trigger print

jrmattila:
#68 

I wanted to programmatically trigger printing and this worked for me.

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -9,6 +9,8 @@ export interface ITriggerProps<T> {
 export interface IReactToPrintProps {
     /** Trigger action used to open browser print */
     trigger: <T>() => React.ReactElement<ITriggerProps<T>>;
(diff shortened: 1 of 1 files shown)


MatthewHerbst:
Hi, thanks for taking a stab at this. While this works for your use case, I don't think it's the right path to take for the library as a whole. We should be exporting a library function that performs the print. For example, you would import it by doing something like `import { printContent } from 'react-to-print`. The current proposal has a few drawbacks:

- It bloats the API
- What happens to the reference after `ReactToPrint` has unmounted?
- It doesn't call the lifecycle methods provided by the other props such as `onBeforeGetContent`

We essentially need to break out all the logic from the current component and put it into a function. Should be straightforward to do, just time consuming.

PR #197, opened 2019-12-23
Make "pageStyle" prop may also be a function

sergeyshmakov:
In some cases we need to calculate pageStyle "just before" the print.
Because of this we need to pass this prop like a function.


MatthewHerbst:
Hello, thanks for the PR. Not sure what's going on, but it looks like you rebased improperly, there is tons of code in this PR that does not belong in it.

sergeyshmakov:
@MatthewHerbst Really true. Dont know how it was happen. Just make upstream from original to my fork and create 1 commit, then pr it. Anyway, sure I will change this.

PR #198, opened 2019-12-24
Make "pageStyle" prop may also be a function

sergeyshmakov:
In some cases we need to calculate pageStyle "just before" the print.
Because of this we need to pass this prop like a function.


README.md
```diff
@@ -144,7 +144,7 @@ The component accepts the following props (note: `?` denotes an optional prop):
...
 |  **`onPrintError`**   | `function(errorLocation: string, error: Error)?` | Callback function that will be called if there is a printing error serious enough that printing cannot continue. Currently limited to Promise rejections in `onBeforeGetContent` or `onBeforePrint`. Use this to attempt to print again. `errorLocation` will tell you in which callback the Promise was rejected.                                                                                     |
 | **`removeAfterPrint`** | `boolean?`  | Remove the print iframe after action. Defaults to `false`.                                                                                                 |
-|    **`pageStyle`**    | `string?`   | Override default print window styling                                                                                               |
+|    **`pageStyle`**    | `string?`, `function?`   | Override default print window styling. If `function` then calling in runtime just before the print.                                                                                               |
```
MatthewHerbst, line 147: Should be `string?|function?`

MatthewHerbst (commented):
Sorry for the delayed review. Could you please add an example of when this would be used? Otherwise looks good

sergeyshmakov:
@MatthewHerbst Thanks for review. I will provide all changes in few days.

MatthewHerbst:
@sergeyshmakov I realize we never got this in. Did you want to add an example and then we can merge it?

MatthewHerbst:
Going to include this in #239

PR #200, opened 2020-01-06
Full load images before print

nhan-huynhvan:
Full loading images before print.

### DIFF
--- a/example/index.tsx
+++ b/example/index.tsx
@@ -30,10 +30,9 @@ class Example extends React.Component<Props, State> {
     private renderContent = () => this.componentRef;
     private renderTrigger = () => <button type="button">Print this out!</button>;
     private onBeforeGetContent = () => {
-        this.setState({
-            text: "text changed",
-            isLoading: true
-        });
+        return new Promise((resolve: any) => 
+            this.setState({ text: "text changed", isLoading: true }, resolve)
+        );
     };
 
     setRef = ref => this.componentRef = ref;
--- a/example/index.tsx
+++ b/example/index.tsx
@@ -26,7 +26,7 @@ class Example extends React.Component<Props, State> {
     }
 
     private handleAfterPrint = () => console.log('after print!');
-    private handleBeforePrint = () => this.setState({ isLoading: false})
+    private handleBeforePrint = () => this.setState({ isLoading: false});
     private renderContent = () => this.componentRef;
     private renderTrigger = () => <button type="button">Print this out!</button>;
     private onBeforeGetContent = () => {
--- a/example/index.tsx
+++ b/example/index.tsx
@@ -7,6 +7,7 @@ import "./relativecss/test.css";
 
 interface State {
     text: string;
+    isLoading: boolean;
 }
 
 interface Props {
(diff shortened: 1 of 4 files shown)


example/index.tsx
```diff
@@ -20,20 +21,27 @@ class Example extends React.Component<Props, State> {
...
         };
     }
 
     private handleAfterPrint = () => console.log('after print!');
-    private handleBeforePrint = () => console.log('before print!');
+    private handleBeforePrint = () => this.setState({ isLoading: false});
     private renderContent = () => this.componentRef;
     private renderTrigger = () => <button type="button">Print this out!</button>;
-    private onBeforeGetContent = () => this.setState({text: "text changed"});
+    private onBeforeGetContent = () => {
```
MatthewHerbst, line 32: You'll have to return a Promise here, since `react-to-print` is not aware of when the `setState` call (which can be asynchronous) completes.

```js
return new Promise((resolve) => {
  this.setState({ text: "text changed", isLoading: true }, resolve);
});
```

src/index.tsx
```diff
@@ -125,13 +126,13 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
...
         const contentNodes = findDOMNode(contentEl);
-        const linkNodes = document.querySelectorAll("link[rel='stylesheet']");
+        const linkNodes = document.querySelectorAll("link[rel='stylesheet'], img");
 
         this.linkTotal = linkNodes.length || 0;
         this.linksLoaded = [];
         this.linksErrored = [];
 
-        const markLoaded = (linkNode, loaded) => {
+        const markLoaded = (linkNode: any, loaded: any) => {
```
MatthewHerbst: `loaded` should be a boolean right?

src/index.tsx
```diff
@@ -154,87 +155,94 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
...
-                            domDoc.head.appendChild(newHeadEl);
-                        } else {
-                            console.warn('"react-to-print" encountered a <link> tag with an empty "href" attribute. In addition to being invalid HTML, this can cause problems in many browsers, and so the <link> was not loaded. The <link> is:', node); // tslint:disable-line max-line-length no-console
-                            // `true` because we"ve already shown a warning for this
-                            markLoaded(node, true);
                         }
                     }
                 }
             }
-
```
MatthewHerbst: Nit: please keep this newline

src/index.tsx
```diff
@@ -154,87 +155,94 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
...
             if (this.linkTotal === 0 || copyStyles === false) {
                 this.triggerPrint(printWindow);
             }
         };
-
```
MatthewHerbst, line 235: Nit: please keep this newline

src/index.tsx
```diff
@@ -154,87 +155,94 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
...
-            domDoc.head.appendChild(styleEl);
-
-            if (bodyClass.length) {
-                domDoc.body.classList.add(bodyClass);
-            }
+                const canvasEls = domDoc.querySelectorAll("canvas");
+                for (let i = 0, canvasElsLen = canvasEls.length; i < canvasElsLen; ++i) {
+                    const node = canvasEls[i];
+                    const contentDrawImage = node.getContext('2d');
+                    if (contentDrawImage) {
```
MatthewHerbst, line 180: Good check, thanks

src/index.tsx
```diff
@@ -154,87 +155,94 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
...
+                            if (node.hasAttribute("href") && !!node.getAttribute("href") ||
+                                (node.hasAttribute('src') && !!node.getAttribute('src'))
+                            ) {
+                                const newHeadEl = domDoc.createElement(node.tagName);
+
+                                // node.attributes has NamedNodeMap type that is not an Array and can be
+                                // iterated only via direct [i] access
+                                for (let j = 0, attrLen = node.attributes.length; j < attrLen; ++j) {
+                                    const attr = node.attributes[j];
+                                    if (attr && attr.nodeValue) {
```
MatthewHerbst: This check needs to be more specific. The empty string `""` is a valid input for the value attribute: https://developer.mozilla.org/en-US/docs/Web/API/Element/setAttribute

MatthewHerbst (changes requested):
These changes look great, just a few things to fix up. Sorry for taking so long to respond, hope you had a happy new years!

src/index.tsx
nhan-huynhvan: Please help to check

MatthewHerbst (approved):
Looks good to me. Are you happy with everything?

Before I merge, would you mind squashing the commits into a single commit please?

PR #230, opened 2020-03-13
Print pdf

invious:
Allows user to pass a boolean to download a PDF instead of bringing up a print dialog

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,6 +146,8 @@ The component accepts the following props (note: `?` denotes an optional prop):
 | **`removeAfterPrint`** | `boolean?`  | Remove the print iframe after action. Defaults to `false`.                                                                                                 |
 |    **`pageStyle`**    | `string?`   | Override default print window styling                                                                                               |
 |    **`bodyClass`**    | `string?`   | Class to pass to the print window body                                                                                     |
+|    **`downloadPDF`**    | `boolean?`   | When passed, instead of printing, a PDF is downloaded                                                                                     |
+|    **`documentTitle`**    | `boolean?`   | Filename for the PDF file                                                                                   |
 |    **`suppressErrors`**    | `boolean?`   | When passed, prevents console logging of errors                                                                                     |
 
 ## FAQ
--- a/example/index.tsx
+++ b/example/index.tsx
@@ -48,6 +48,7 @@ class Example extends React.Component<Props, State> {
                     onBeforePrint={this.handleBeforePrint}
                     onAfterPrint={this.handleAfterPrint}
                     removeAfterPrint
+                    downloadPDF
                 />
                 <ComponentToPrint ref={this.setRef} text={this.state.text}/>
             </div>
--- a/package.json
+++ b/package.json
@@ -37,6 +37,8 @@
     "react-dom": "^15.0.0 || ^16.0.0"
   },
   "dependencies": {
+    "downloadjs": "^1.4.7",
+    "js-base64": "^2.5.2",
     "prop-types": "^15.7.2"
   },
   "devDependencies": {
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -48,11 +48,11 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
 
     public startPrint = (target: any, onAfterPrint: any) => {
         const { removeAfterPrint } = this.props;
-        const docAsBase64 = templateToUri(target.contentWindow.document.documentElement.outerHTML)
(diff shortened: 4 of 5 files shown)


src/index.tsx
```diff
@@ -29,19 +31,34 @@ export interface IReactToPrintProps {
...
+
 export default class ReactToPrint extends React.Component<IReactToPrintProps> {
     private linkTotal: number;
     private linksLoaded: Element[];
     private linksErrored: Element[];
 
     public startPrint = (target: any, onAfterPrint: any) => {
-        const { removeAfterPrint } = this.props;
+        const { removeAfterPrint, downloadPDF, documentTitle } = this.props;
+        const docAsBase64 = templateToUri(target.contentWindow.document.documentElement.outerHTML);
```
chefboyarsky: Looks like call to templateToUri is only needed in downloadPDF=true case and can be moved into condition on line 56

src/index.tsx
```diff
@@ -29,19 +31,34 @@ export interface IReactToPrintProps {
     removeAfterPrint?: boolean;
     /** Optional - suppress error messages */
     suppressErrors?: boolean;
+    /** Optional - download pdf file directly */
+    downloadPDF?: boolean;
+    /** Optional - document title for PDF file */
+    documentTitle?: string;
 }
 
+export const templateToUri = (html) => (
```
chefboyarsky, line 40: Docblock

example/index.tsx
```diff
@@ -48,6 +48,7 @@ class Example extends React.Component<Props, State> {
                     onBeforePrint={this.handleBeforePrint}
                     onAfterPrint={this.handleAfterPrint}
                     removeAfterPrint
+                    downloadPDF
```
MatthewHerbst, line 51: I think we should create a new example for this, since this isn't the default behavior but it makes the example use the behavior.

README.md
```diff
@@ -146,6 +146,8 @@ The component accepts the following props (note: `?` denotes an optional prop):
 | **`removeAfterPrint`** | `boolean?`  | Remove the print iframe after action. Defaults to `false`.                                                                                                 |
 |    **`pageStyle`**    | `string?`   | Override default print window styling                                                                                               |
 |    **`bodyClass`**    | `string?`   | Class to pass to the print window body                                                                                     |
+|    **`downloadPDF`**    | `boolean?`   | When passed, instead of printing, a PDF is downloaded                                                                                     |
+|    **`documentTitle`**    | `boolean?`   | Filename for the PDF file                                                                                   |
```
MatthewHerbst, line 150: Can we use this to also modify the print window title (line 139)?

package.json
```diff
@@ -37,6 +37,8 @@
     "react-dom": "^15.0.0 || ^16.0.0"
   },
   "dependencies": {
+    "downloadjs": "^1.4.7",
```
MatthewHerbst, line 40: I'm not particularly happy with this since it's not maintained and doesn't have much of a community. Let's use [`file-saver`](https://www.npmjs.com/package/file-saver) instead please. Let me know if you have any trouble using it

src/index.tsx
```diff
@@ -29,19 +31,36 @@ export interface IReactToPrintProps {
     removeAfterPrint?: boolean;
     /** Optional - suppress error messages */
     suppressErrors?: boolean;
+    /** Optional - download pdf file directly */
+    downloadPDF?: boolean;
+    /** Optional - document title for PDF file */
+    documentTitle?: string;
```
MatthewHerbst, line 37: What is the title of the PDF if this is not passed? Should there be a default value to the prop?

src/index.tsx
```diff
@@ -29,19 +31,36 @@ export interface IReactToPrintProps {
...
 }
 
+export const templateToUri = (html) => (
```
MatthewHerbst, line 40: Maybe `htmlToBase64Uri` instead? Not sure what `template` refers to

MatthewHerbst (changes requested):
This will be a cool feature! Let's just use the more modern/supported library and then we should be good to go

MatthewHerbst:
I'm also not sure you've actually done conversion to PDF: just downloading HTML. You might want to look into this library instead (which uses `file-saver` under the hood): https://www.npmjs.com/package/jspdf

MatthewHerbst:
That's a really big library though, so Ideally it can be avoided if possible

MatthewHerbst:
Hey everyone, what's the plan going forward with this PR?

Issue #219, opened 2020-02-22
ReactToPrint trigger by ref of button outside component


soundquiet:
hi, I want to know if it is possible to use `ref` for `trigger`. I would like to put the print button in the parent component.

I tried `trigger={()=>this.printRef}` but it did not work.

Did I make some misunderstandings? thanks a lot.

MatthewHerbst:
Hello. This is not currently possible, however, it could probably be made possible with a relatively simple PR. I'm currently traveling so can't make one myself. I'd be happy to review one of you can make it.

MatthewHerbst:
This is available in `2.8.0-beta.1` and will be available in `2.8.0` later tonight. Thanks fo the ideas!

PR #244, opened 2020-05-05
Feature/more controls over custom trigger

vtsybulin:
@gregnb @MatthewHerbst  Hello! Faced kinda same need as per https://github.com/gregnb/react-to-print/issues/219

However I need even more control then just binding to some single trigger ref, so I decided to go with context provider to expose the `handleClick` method directly to be used afterward. Sample:
```
import ReactToPrint, { PrintContext } from 'react-to-print';

...

<ReactToPrint content={() => this.componentRef}>
    <PrintContext.Consumer>
	{({ handlePrint }) => (
          <button onClick={handlePrint}>Print this out!</button>
	)}
     </PrintContext.Consumer>
</ReactToPrint>
```

Also wrapped ReactToPrint into [useReactToPrint] hook to be used with functional components. Sample:
```
import { useReactToPrint } from 'react-to-print';

...

const Example = () => {
  const componentRef = useRef();
  const handlePrint = useReactToPrint({
    content: () => componentRef.current,
  });
  
  return (
    <div>
      <ComponentToPrint ref={componentRef} />
      <button onClick={handlePrint}>Print this out!</button>
    </div>
  );
};
```

Let me know what you think! Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,12 +51,12 @@ The component accepts the following props:
 |    **`bodyClass`**    | `string?`   | Class to pass to the print window body                                                                                     |
 |    **`suppressErrors`**    | `boolean?`   | When passed, prevents `console` logging of errors
 
-### PrintContext
+### PrintContextConsumer
 
-In case you don't want to specify `trigger` directly within other props but you rather want to gain some extra control over it, you could make use of PrintContext. By using PrintContext you can get direct access to the `handlePrint` method which triggers printing action.
+If you need extra control over printing and don't want to specify `trigger` directly, `PrintContextConsumer` allows you to gain direct access to the `handlePrint` method which triggers the print action. Requires React 16.3.0 and higher.
 
 ### [useReactToPrint]
-For the functional components, there is a useReactToPrint hook, which accepts an object with the same configuration props as `<ReactToPrint />` itself and returns a handlePrint function, by calling which you can trigger printing action.
+For functional components, use the `useReactToPrint` hook, which accepts an object with the same configuration props as `<ReactToPrint />` returns a `handlePrint` function which when called will trigger the print action. Requires React 16.8 and higher.
  
 ## Example
 
@@ -102,10 +102,10 @@ class Example extends React.Component {
 }
 ```
 
-### Calling from class components with PrintContext
+### Calling from class components with PrintContextConsumer
 ```js
 import React from 'react';
-import ReactToPrint, { PrintContext } from 'react-to-print';
+import ReactToPrint, { PrintContextConsumer } from 'react-to-print';
 
 class ComponentToPrint extends React.Component {
   render() {
@@ -135,11 +135,11 @@ class Example extends React.Component {
         <ReactToPrint
           content={() => this.componentRef}
         >
-        	<PrintContext.Consumer>
+        	<PrintContextConsumer>
 				{({ handlePrint }) => (
 					<button onClick={handlePrint}>Print this out!</button>
 				)}
-			</PrintContext.Consumer>
+			</PrintContextConsumer>
         </ReactToPrint>
         <ComponentToPrint ref={el => (this.componentRef = el)} />
       </div>
@@ -149,7 +149,7 @@ class Example extends React.Component {
 ```
 
 
-### Calling from functional components with [hooks](https://reactjs.org/docs/hooks-intro.html)
+### Calling from functional components with `useReactToPrint` [hook](https://reactjs.org/docs/hooks-intro.html)
 
 ```js
 import React, { useRef } from 'react';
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -51,6 +51,13 @@ The component accepts the following props:
 |    **`bodyClass`**    | `string?`   | Class to pass to the print window body                                                                                     |
 |    **`suppressErrors`**    | `boolean?`   | When passed, prevents `console` logging of errors
 
+### PrintContext
+
+In case you don't want to specify `trigger` directly within other props but you rather want to gain some extra control over it, you could make use of PrintContext. By using PrintContext you can get direct access to the `handlePrint` method which triggers printing action.
```
MatthewHerbst:
```suggestion
If you need extra control over printing and don't want to specify `trigger` directly, `PrintContext` allows you to gain direct access to the `handlePrint` method which triggers the print action.
```

README.md
```diff
@@ -51,6 +51,13 @@ The component accepts the following props:
...
+
+### [useReactToPrint]
+For the functional components, there is a useReactToPrint hook, which accepts an object with the same configuration props as `<ReactToPrint />` itself and returns a handlePrint function, by calling which you can trigger printing action.
```
MatthewHerbst:
```suggestion
For functional components, use the `useReactToPrint` hook, which accepts an object with the same configuration props as `<ReactToPrint />` returns a `handlePrint` function which when called will trigger the print action.
```

README.md
```diff
@@ -135,6 +189,48 @@ const Example = () => {
     </div>
   );
 };
+
+```
+### Calling from functional components with [useReactToPrint](https://reactjs.org/docs/hooks-intro.html)
```
MatthewHerbst, line 194:
```suggestion
### Calling from functional components with `useReactToPrint` [hook](https://reactjs.org/docs/hooks-intro.html)
```

src/index.tsx
```diff
@@ -1,6 +1,11 @@
 import * as React from "react";
 import { findDOMNode } from "react-dom";
 
+export interface IPrintContextProps {
+    handlePrint: () => void,
+};
+export const PrintContext = React.createContext({} as IPrintContextProps);
```
MatthewHerbst: Is there a way to make this conditional based on `React.createContext` existing? Right now we support React v15 which does not have `createContext` (introduced in React `16.3.0`). I would actually be ok making a major version change to drop React 15, however, I'd want to get a few more breaking changes in I've been meaning to do, which would delay shipping this by a few days.

src/index.tsx
```diff
@@ -322,11 +327,23 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
...
+        } else {
+            const value = {handlePrint: this.handleClick};
+            return <PrintContext.Provider value={value as IPrintContextProps}>
+                {children}
+            </PrintContext.Provider>
+        }
     }
 }
+
+export const useReactToPrint = (options: IReactToPrintProps) => {
```
MatthewHerbst: Similar to the above comment, is there a way to make this a no-op (just export `undefined` or something) if the user is on React 15 (which doesn't have `useMemo` and `useCallback`?

MatthewHerbst (commented):
This is fantastic, thank you!

My only concern is that it would be a breaking change since we currently support React 15. As mentioned in the review comments, I don't mind making that breaking change, but would want to add a few more breaking changes in at the same time which could delay this by a couple of days. Do you mind waiting until the weekend for this to be released? Happy to get it out sooner if you can make the code not break when using React 15.

vtsybulin:
@MatthewHerbst Hi. Thanks for the feedback! Sounds cool. Sorry for the v15 thing missed that :/ Let me take a look tomorrow so I can update accordingly If you're ok with that?

MatthewHerbst:
No problem. If you're not in a rush then I'm happy to move deploying this to the weekend so I can get a few other changes in, and then we can do a major version release and drop React 15 support (which we should do anyways for a variety of reasons).

jtart:
hey @MatthewHerbst, are you close to releasing this at all?

vtsybulin:
@MatthewHerbst Hey! Sorry, it's been a while, was a bit busy. I hope it's better now. Let me know if anything extra is required, thanks!

vtsybulin:
@MatthewHerbst nice! thanks 👍

MatthewHerbst:
@vtsybulin thank you! I'm working on a few small things, will have it in a new release in a couple of hours

vtsybulin:
@MatthewHerbst Cool. Let me know if you have anything else in mind to assist with.

MatthewHerbst:
This is published with the just released version `2.8.0-beta.1`. I'll release `2.8.0` either later tonight or tomorrow once I have a chance to do more testing.

MatthewHerbst:
Published as `2.8.0`, thanks again!

vtsybulin:
Great 👍

PR #245, opened 2020-05-06
Feature/ new documentTitle props to set filename for pdf print

### LINKED ISSUE #243: default title when print to pdf.
Hello,

Actually when i try to print to pdf, it takes page's title defined in the head tag.

Is there any existing props to customize page title for print to pdf ? Anyway, it could be a great feature.

Thanks

EDIT : i've seen there was another issue about this. #122 . But there was not real solution, just workarounds. My problem is i'm using Helmet to update document title dynamically. And i don't want to change it in browser tab when printing. for my pdf, I just need to set a different filename from document title.

EDIT2 : i've found a workaround. basically i change document.title beforePrint and reset with previous saved title afterPrint.
 But it is not very clean because browser tab title really change before and after. Also, it brings some complexity for this simple feature.

Example to Change document title before and after : 
```
<ReactToPrint
            trigger={...}
            content={() => printable.current}
             onBeforePrint={() => document.title = "Print page title"}
             onAfterPrint={() => document.title = originalTitle}
/>
```

Complexity to save original title with useState and useEffect:
```
const [originalTitle, setOriginalTitle] = useState('');

useEffect(() => {
  
       setOriginalTitle(printable.current.ownerDocument.title) 
       //here not just document.title because  document.title take the value of <title> tag.
       //But with Helmet real title could be different
       //So we want the real Helmet updated title and this one is in printable.current.ownerDocument
       
}, [printable]);
```

zb2oby:
Here a PR to add ability to set custom filename when print to pdf file with a new props documentTitle.
It causes a temporary override of the document.title (no other way to do this, print api does not provide a way to ) and reset it after print.

Fixes #243

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -9,6 +9,8 @@ export interface ITriggerProps<T> {
 type PropertyFunction<T> = () => T;
 
 export interface IReactToPrintProps {
+    /** Optional title for document if saved as file */
+    documentTitle?: string
     /** Class to pass to the print window body */
     bodyClass?: string;
     /** Content to be printed */
@@ -50,6 +52,7 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
             onAfterPrint,
             removeAfterPrint,
             suppressErrors,
+            documentTitle,
         } = this.props;
 
         setTimeout(() => {
@@ -59,7 +62,19 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
                 // Some browsers, such as Firefox Android, do not support printing at all
                 // https://developer.mozilla.org/en-US/docs/Web/API/Window/print
                 if (target.contentWindow.print) {
+
+                    // save current document title before override
+                    const tempTitle = document.title;
+                    if (documentTitle) {
+                        document.title  = documentTitle; // It causes overriding tab title during the print process
+                    }
+
                     target.contentWindow.print();
+
+                    // reset document title with original title after print process
+                    if (documentTitle) {
+                        document.title = tempTitle;
+                    }
 
                     if (onAfterPrint) {
                         onAfterPrint();
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -9,14 +9,14 @@ export interface ITriggerProps<T> {
 type PropertyFunction<T> = () => T;
 
 export interface IReactToPrintProps {
-    /** Optional title for document if saved as file */
-    documentTitle?: string
     /** Class to pass to the print window body */
     bodyClass?: string;
     /** Content to be printed */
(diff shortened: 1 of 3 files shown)


src/index.tsx
```diff
@@ -59,8 +62,20 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
                 // Some browsers, such as Firefox Android, do not support printing at all
                 // https://developer.mozilla.org/en-US/docs/Web/API/Window/print
                 if (target.contentWindow.print) {
+
+                    // save current document title before override
+                    const tempTitle = document.title;
+                    if (documentTitle) {
+                        document.title  = documentTitle; // It causes overriding tab title during the print process
```
MatthewHerbst:
```suggestion
                        document.title  = documentTitle; // Overrides the tab title during the print process
```

src/index.tsx
```diff
@@ -59,8 +62,20 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
                 // Some browsers, such as Firefox Android, do not support printing at all
                 // https://developer.mozilla.org/en-US/docs/Web/API/Window/print
                 if (target.contentWindow.print) {
+
+                    // save current document title before override
```
MatthewHerbst:
```suggestion
```

src/index.tsx
```diff
@@ -59,8 +62,20 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
...
+                    const tempTitle = document.title;
+                    if (documentTitle) {
+                        document.title  = documentTitle; // It causes overriding tab title during the print process
+                    }
+
                     target.contentWindow.print();
 
+                    // reset document title with original title after print process
```
MatthewHerbst:
```suggestion
```

src/index.tsx
```diff
@@ -9,6 +9,8 @@ export interface ITriggerProps<T> {
 type PropertyFunction<T> = () => T;
 
 export interface IReactToPrintProps {
+    /** Optional title for document if saved as file */
+    documentTitle?: string
```
MatthewHerbst: Nit: please alphabetize (should come after `copyStyles`)

MatthewHerbst (commented):
Great, thank you! Can you please also add the prop to the README?

src/index.tsx
zb2oby: alphabetized in commit e2a7c3d6f855d731c60f3cf621efdf46fbd0ca54

MatthewHerbst:
This is published with the just released version `2.8.0-beta.1`. I'll release `2.8.0` either later tonight or tomorrow once I have a chance to do more testing.

Issue #254, opened 2020-06-02
Clear onload in all browsers (not just IE)?


JoshuaKGoldberg:
Hi! This library is great, thanks so much for making it! 💖 

Is there a need to  [only set `printWindow.onload = null;` in IE](https://github.com/gregnb/react-to-print/blob/a192abd07cda318b89d7eaf81198c985cde58b97/src/index.tsx#L228)? Seems like that could run regardless of browser right?

https://github.com/gregnb/react-to-print/blob/a192abd07cda318b89d7eaf81198c985cde58b97/src/index.tsx#L228

Some backing context for why I'd like to remove that line: https://codesandbox.io/s/react-to-print-infinite-onloads-h4mnr infinitely hangs when running tests.
* Enzyme continuously calls the [`printWindow`'s `onload`](https://github.com/gregnb/react-to-print/blob/a192abd07cda318b89d7eaf81198c985cde58b97/src/index.tsx#L226), even though it's only been set once.
* That causes the `act` in the test to never complete...
  * ...but the `onBeforeLoad` function never gets called if you use a synchronous `act`.

You can verify this by cloning the sandbox locally, prettifying the compiled `index.js`, and adding console logs.

```js
console.log("Setting onload");
f.onload = function() {
  console.log("Calling onload");
    window.navigator && window.navigator.userAgent.indexOf("Trident/7.0") > -1 && (f.onload = null);
```

```
console.log
    Setting onload
      at e.handlePrint (node_modules/react-to-print/lib/index.js:109:35)

  console.log
    Calling onload
      at HTMLIFrameElement.f.onload (node_modules/react-to-print/lib/index.js:111:37)

  console.log
    Calling onload
      at HTMLIFrameElement.f.onload (node_modules/react-to-print/lib/index.js:111:37)

  console.log
    Calling onload
      at HTMLIFrameElement.f.onload (node_modules/react-to-print/lib/index.js:111:37)
...
```

Now, this seems like a bug in Enzyme and should probably be fixed there (https://github.com/enzymejs/enzyme/issues/2400), _but_ it would be fixed by this already-seemingly-positive change.

MatthewHerbst:
Hi, thanks for the report. That does seem like an interesting issue. And I agree, I don't see why we can't just always clear out the event listener. I'll get it in with some changes I'm making tonight!

MatthewHerbst:
Should be fixed in the just published `2.9.0`, let me know if you see any issues.

[...]

PR #258, opened 2020-06-03
Converted from TSLint to ESLint

JoshuaKGoldberg:
Continues the discussion from #254 by:

1. Replacing `tslint.json` and its related npm dependencies with `.eslintrc.js` and its related npm dependencies
2. Disabling the ESLint rules obviously not desirable in this project, such as `@typescript-eslint/no-explicit-any`
3. Copying the `max-len` settings over to the new config
4. Removing any violations of rules that seemed ok to remove, such as the unnecessary `as` cast


src/index.tsx
```diff
@@ -254,7 +254,7 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
                     const node = canvasEls[i];
                     const contentDrawImage = node.getContext("2d");
                     if (contentDrawImage) {
-                        contentDrawImage.drawImage(srcCanvasEls[i] as HTMLCanvasElement, 0, 0);
+                        contentDrawImage.drawImage(srcCanvasEls[i], 0, 0);
```
MatthewHerbst, line 257: Was there a linting issue here?

MatthewHerbst (commented):
Fantastic, thank you! Just one question about a type change that was made

src/index.tsx
JoshuaKGoldberg: Yup, the `as` cast was unnecessary. The element was already that type.

JoshuaKGoldberg:
> changed the title

Dangit, I always get this wrong... 😄

MatthewHerbst:
Awesome, thanks again!

PR #266, opened 2020-07-03
Fix `

Domino987:


MatthewHerbst:
Hi, thanks, but this is actually a big with github not showing the right thing, not an issue with the markdown formatting. Putting a `|` between the two values causes `function` to become the value for the description column.

PR #267, opened 2020-07-09
README fix: escape pipe in pageStyles string | function

JoshuaKGoldberg:
Pipes are used as table column markers so they need to be escaped for inline code blocks. ✨


MatthewHerbst:
Hi, thanks for looking at this. Question: is our MD wrong, or is GitHub not displaying the right thing? My assumption is that because `string | function` is wrapped as a code block, that everything inside of it should be parsed normally, and not be considered special syntax.

Related previous PR: https://github.com/gregnb/react-to-print/pull/266

JoshuaKGoldberg:
Maybe GitHub is the off one here? 🤷

MatthewHerbst:
> Maybe GitHub is the off one here? 🤷

Yeah, that's my thinking, but I would be happy to be proved wrong. I should do a bigger dive into the MD spec to see what (if anything) it says about this. Maybe there is an existing GitHub issue for this?

MatthewHerbst:
https://github.github.com/gfm/#code-spans

Specifically regarding your commit:

> Note that backslash escapes do not work in code spans. All backslashes are treated literally:

MatthewHerbst:
Here is the table part of the spec: https://github.github.com/gfm/#tables-extension-

And relevant to us:

> Each row consists of cells containing arbitrary text, in which inlines are parsed, separated by pipes (|).

As seen in the code span part of the spec, code spans are inlines, so, I think GitHub has this one wrong, unless I'm reading the spec incorrectly

MatthewHerbst:
Upstream tracking: https://github.com/gjtorikian/commonmarker/issues/125

JoshuaKGoldberg:
😂 . It's always nice to see an upstream bug filed on common tooling for a PR... thank you!

MatthewHerbst:
I wonder if we should still modify our README somehow for now to make the documentation not be broken here, though npm properly displays the README so maybe it's not an issue, since I imagine most people will read it there, not here. We could write out `or` instead of using the `|`?

README.md
```diff
@@ -48,7 +48,7 @@ The component accepts the following props:
...
 | **`removeAfterPrint?`** | `boolean`  | Remove the print iframe after action. Defaults to `false`.                                                                                                 |
-|    **`pageStyle?`**    | `string | function`   | We set some basic styles to help improve page printing. Use this to override them and provide your own. If given as a function, it must return a `string`                                                                                               |
+|    **`pageStyle?`**    | `string \| function`   | We set some basic styles to help improve page printing. Use this to override them and provide your own. If given as a function, it must return a `string`                                                                                               |
```
MatthewHerbst: From the most recent reply in the upstream thread, it seems the current behavior is correct actually, and npm and the others have it wrong. I still think most people are going to be reading the README on npm, but we should for sure have a solution that works on both.

It's a bit hacky, but maybe let's change this to `string or function` just to get around the problem for now? Happy to hear a different solution if you think of something better

MatthewHerbst (changes requested):

README.md
JoshuaKGoldberg: Sure, I'm ok with that :+1:

Thinking "out loud": `|` is closer to how TypeScript would represent the types... but not everyone would understand what `|` means, so I think an `or` works well.

PR #269, opened 2020-07-20
README Page Break Guideines

hbrannan:
First pass at generic solution to page-breaking React content.


MatthewHerbst (approved):
This is great, thank you!

PR #274, opened 2020-07-28
Print replacer example for Electron (silent printing)

ririshi:
A use-case for the `print` replacement callback. It uses Electron's printing functionality, with a workaround to allow Electron to "print an iframe" by creating an invisible window that copies the contents of react-to-print iframe. This window sends a print request to the Electron main process through Electron's IPC messaging. The main process will attempt to silently print the window, and sends the result of this effort back to the window in the form of a Promise. This Promise is then returned by the `print` callback to allow react-to-print to continue.

### DIFF
--- a/example/ElectronPrint/electron/electron.js
+++ b/example/ElectronPrint/electron/electron.js
@@ -1,0 +1,96 @@
+const {
+  app, BrowserWindow, protocol, ipcMain, webContents
+} = require('electron');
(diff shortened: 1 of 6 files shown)


MatthewHerbst:
Working Electron example with how to print using `react-to-print` is now available! https://stackoverflow.com/questions/68751490/electron-print-iframe-given-reference-to-it

Issue #295, opened 2020-10-15
bodyClass property does not allow for multiple style class names


seanblonien:
I wanted to pass multiple CSS classes to the ReactToPrint `bodyClass` property, but it only accepts a string with one class name. The expected behaviour would be to allow for more than one class.

I believe the line of code causing this problem is from the [classList.add](https://github.com/gregnb/react-to-print/blob/28365f179aa181dd489c2d99484673eb99cf7684/src/index.tsx#L255) method.

Here's the [codesandbox ](https://codesandbox.io/s/silly-pascal-66nxn?file=/src/App.js) of the desired behaviour, but clicking on React to print will cause an `InvalidCharacterError`.

PR #296, opened 2020-10-15
Split bodyClass string to allow multiple body classes

seanblonien:
From issue #295, simply allow for a bodyClass string with multiple class names and spread them as arguments for `classList.add` method.

Passing this prop would throw an error and the print would never render.
```
bodyClass="firstClass secondClass"
```

With this fix, you are allowed to pass more than 1 class names without error.


MatthewHerbst:
Awesome, thank you!

MatthewHerbst:
Published as v`2.10.1`, thanks again!

Issue #66, opened 2018-11-02
Print with check box checked


jsbhakta:
I have been using this library and found an issue not sure if some one has noticed it or if i am missing something to print checkbox. but it seems when the print command executed it removed the checked sign out of checkbox.

__page with checkbox__
<img width="1435" alt="screen shot 2018-11-01 at 6 19 24 pm" src="https://user-images.githubusercontent.com/19602461/47888634-c9b23380-de02-11e8-9073-b474be9c5994.png">

__Actual Print preview__
<img width="1432" alt="screen shot 2018-11-01 at 6 19 38 pm" src="https://user-images.githubusercontent.com/19602461/47888639-cfa81480-de02-11e8-9294-4875cb9b4367.png">

dsbhakta:
Yes very good finding. I thought i was doing something wrong. good examples.

ktulchinsky:
Super awesome, it should support checbox checked print

purevsurenm:
I also see this issue. Not sure why. It is confusing when you fill up a form and print it, but the checkboxes you checked are unchecked.

MatthewHerbst:
This is a very strange bug. Does anyone know of any other circumstances where similar behavior is seen? I'm trying to narrow down where I should be looking to solve this. Thanks for the reports!

mdland-service:
Any news on the this? Experiencing the same issue - only display checkboxes that have defaultValues

valeriozhang:
Any news on the this? Experiencing the same issue - only display checkboxes that have defaultValues

MatthewHerbst:
@valeriozhang I've not been able to make any progress on this. My guess is it's something arcane. Do you have any ideas?

kapilgundawar:
Hi Team,
iam also facing checkbox issue the moment i click on print,All my checked are removed. 
Can someone please help in this

valeriozhang:
@MatthewHerbst No sorry - haven't looked through the source. I just used CSS to modify my browser prints.

MatthewHerbst:
@valeriozhang what CSS did you add?

valeriozhang:
I didnt use the package.

```css
@media print {
    body * {
      visibility: hidden;
    }
  
    #section-to-print, #section-to-print * {
      visibility: visible;
    }
  
    #section-to-print {
      /* position: absolute;
      left: 0;
      top: 0;
      bottom: 200 */
    }
    
}
```

ibanjb:
[...]

PR #298, opened 2020-10-16
Fixing checkboxes state not copied - #66

aviklai:
Fixing checkboxes state not being copied to the print window


MatthewHerbst:
I've verified locally that this works, thanks! It's unfortunate that we need to do this at all since it would have been nice if the browser just copied the state for us.

MatthewHerbst:
Published as v`2.10.2`, thanks 🎉

PR #301, opened 2020-10-16
Use Object.prototype.hasOwnProperty. Resolves #300.

### LINKED ISSUE #300: react-to-print is incompatible with SystemJS (and ES modules)
react-to-print assumes that the `React` object has `Object.prototype` as its prototype, since it calls `React.hasOwnProperty()` and the `hasOwnProperty` function only exists on the Object.prototype.

https://github.com/gregnb/react-to-print/blob/dab3e2224b526f57130615fb977c42359ff10317/src/index.tsx#L4-L5

However, this assumption is not true of ES modules, nor of SystemJS modules. The ES module spec says that ES modules must have a prototype of `null`, which means that they do not have the `hasOwnProperty` function available on them. This issue has been discussed and fixed in a variety of places, including the following links:

- https://github.com/systemjs/systemjs/issues/2147#issuecomment-600880479
- https://github.com/tannerlinsley/react-table/pull/2021
- https://github.com/mosch/react-avatar-editor/pull/339
- https://github.com/rollup/rollup/pull/3420

I help maintain SystemJS, and a user reported this error originally with us (see https://single-spa.slack.com/archives/C8R6U7MT7/p1602879935124300).

The following code sandbox shows how the `react-to-print` library dies if provided a React object without Object.prototype as the prototype: https://codesandbox.io/s/blissful-flower-wy5k7?file=/index.html

The fix in this case is to call `Object.prototype.hasOwnProperty.call(React, 'useContext')`. I'll submit a PR with that fix.

internettrans:
See #300

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -1,8 +1,8 @@
 import * as React from "react";
 import { findDOMNode } from "react-dom";
 
-const contextEnabled = React.hasOwnProperty("createContext");
-const hooksEnabled = React.hasOwnProperty("useMemo") && React.hasOwnProperty("useCallback");
+const contextEnabled = Object.prototype.hasOwnProperty.call(React, "createContext");
+const hooksEnabled = Object.prototype.hasOwnProperty.call(React, "useMemo") && Object.prototype.hasOwnProperty.call(React, "useCallback");
 
 export interface IPrintContextProps {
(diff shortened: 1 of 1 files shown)


rkgummadi:
thnx @joeldenning  for fixing the issue

PR #537, opened 2022-10-19
Fix disabled node breaking the print

luckrnx09:
Print is not work if there is any disabled link node in the source content element.

Example:
here is a content element ready for print
```jsx
<div>
    <link rel="stylesheet" href="my-href-one"></link>
    <link disabled rel="stylesheet" href="my-href-two">
    </link>
    <div>
        My Content To Print
    </div>
</div>
```
A `disabled` attribute is set to the second link node, it never trigger `onload` or `onerror` callbacks after append it into an iframe.
FYI: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#attr-disabled
So we can mark them loaded directly, otherwise, the `numResourcesManaged` is not equals to `this.linkTotal` in the `handlePrint` function
```tsx
  // numResourcesManaged always equals 1 and this.linkTotal always equals 2 in this case above
  if (numResourcesManaged === this.linkTotal) {
       this.triggerPrint(printWindow);
  }
```

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -444,8 +444,8 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
                             // the current page. Some will attempt to load the page"s parent
                             // directory. These problems can cause `react-to-print` to stop without
                             // any error being thrown. To avoid such problems we simply do not
-                            // attempt to load these links.
-                            if (node.getAttribute("href")) {
+                            // attempt to load these links. 
+                            if (node.getAttribute("href") && !node.hasAttribute("disabled")) {
                                 const newHeadEl = domDoc.createElement(node.tagName);
 
                                 // Manually re-create the node


luckrnx09:
@cmckni3  I found this bug in my production environment. Could please review and release a new version to fix this problem if you have chance. thank you very much!

src/index.tsx
```diff
@@ -444,8 +444,8 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
                             // the current page. Some will attempt to load the page"s parent
                             // directory. These problems can cause `react-to-print` to stop without
                             // any error being thrown. To avoid such problems we simply do not
-                            // attempt to load these links.
-                            if (node.getAttribute("href")) {
+                            // attempt to load these links. 
+                            if (node.getAttribute("href") && !node.hasAttribute("disabled")) {
```
luckrnx09, line 448: any link node with `disabled`attribute, regardless of any of the following, will not trigger `onload` or `onerror`
 callback. We should skip them.
```
<link disabled ... /> 
<link disabled="true" ... />`
<link disabled="false" ... />` 
```

luckrnx09:
@cmckni3 @MatthewHerbst 
I found this problem in my production environment. PTAL for this PR and release a new version to fix it. Thank you!

MatthewHerbst:
Hi, thanks a lot for raising this issue and for submitting a PR, it's super appreciated. I'm having a hard time consistently reproducing the issue, though I have been able to see it inconsistently. I'm working on finding a way to consistently reproduce it so I can add a test to the examples to ensure we son't run into something like this again. Hopefully can get this merged tomorrow, thank you!

luckrnx09:
hi @MatthewHerbst , Is this more complicated than we think? If you need me to provide the reproduction steps, such as screen recording, please let me know. :)

MatthewHerbst:
Hi, apologies. If you could provide consistent reproduction steps that would be helpful, I'm still only able to get it intermittently (which itself isn't a problem, but I want to ensure this fix doesn't break anything for users who may not be seeing the issue: I don't think it does, just like to be safe).

I also just moved this past week so I haven't been looking at the computer much, apologies. I will take a look at this again this week. If you don't hear from me by EOD Wednesday please bump again.

luckrnx09:
@MatthewHerbst Of course, I created a demo for reproduce the problem
[codesandbox demo](https://codesandbox.io/s/goofy-bell-hfcp6u?file=/src/App.js)
You can see my [screen recording](https://filetransfer.io/data-package/EHj2cLI8#link) for it

MatthewHerbst:
Something interesting I've discovered: there is absolutely no issue with using disabled `link`s when you put them directly into the HTML. It's specifically when putting a disabled `link` into the JSX that the problem appears. 

I'd like to learn more about the above, but for now I'm happy that we can consistently reproduce

MatthewHerbst:
Going to merge this so you get credit for it, then make a small PR to add a test for it/clean up the warning message it produces.

Thanks for figuring this out!

MatthewHerbst:
@luckrnx09 please let me know what you think of this PR: https://github.com/gregnb/react-to-print/pull/543

MatthewHerbst:
The changes have been published in v2.14.8, thanks again for reporting this and for the help here!

PR #543, opened 2022-11-02
Test and cleanup the `disabled` <link> node logic

MatthewHerbst:
Follow up to #537 

@luckrnx09 Let me know what you think!


luckrnx09:
@MatthewHerbst  Thank you for your careful consideration. I need to spend some time to study whether it can be filtered through the selector. I think I will reply to you tomorrow

luckrnx09:
hi @MatthewHerbst , sorry for reply you so late.
Almost looks good for me, but two ideas about this PR: 
1. Prefer add a `rel="stylesheet"` for each `link` tags you added in this PR, because we are importing the css files from outside.
2. We can use a selector `link["stylesheet"]:not([disabled])` to filter links which has the `disabled` attribute, because the css will not be applied if the link has attribute `disabled="false"`.
See: https://codesandbox.io/s/zealous-http-0g0p4c?file=/index.html

Correct me If I am wrong

luckrnx09:
correct me If I am wrong

luckrnx09:
correct me If I am wrong

src/index.tsx
```diff
@@ -444,30 +445,43 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
...
-                                        newHeadEl.setAttribute(attr.nodeName, attr.nodeValue || "");
+                            // attempt to load these links.
+                            if (node.getAttribute("href")) {
+                                // Browser's don't display `disabled` `link` nodes, so we need to filter them out
+                                // https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#attr-disabled
+                                // https://caniuse.com/mdn-html_elements_link_disabled
+                                // TODO: ideally we could just filter these out on selection, by
+                                // using a selector such as: `link[rel='stylesheet']:not([disabled=true][disabled=false])`
+                                // https://stackoverflow.com/questions/27733826/css-selectors-for-excluding-by-attribute-presence
+                                // However, that doesn't seem to work. Why?
```
luckrnx09, line 456: In the example above, we added many `link` tags but missing the `rel="stylesheet"` attribute, so the selector can not work well.

MatthewHerbst:
I haven't been able to get the selector to work for some reason. We can revisit it in the future if we figure it out. For now I think this is good to go, and I made some improvements to the examples (tests 😉 ) as well

PR #553, opened 2022-11-15
Fix viewport not being set correctly

### LINKED ISSUE #327: Printing uses mobile viewport instead of desktop
Hi ,

In my application I applied styling for the content based on the view port. For example for the desktop I am using col-lg-3 , for ipad i am using col-md-3 and for mobile- col-sm-6 and col-xs-6. But on the print preview it is taking only mobile version styling sm and xs even on when the user is clicking on print from desktop. Is there any fix for this issue where it supports lg and md as well?

MatthewHerbst:
Fixes https://github.com/gregnb/react-to-print/issues/327

Without this the browser sets the `iframe` to its default size of width 300px x length 150px, which could cause the browser to print the components differently if there are styles present that change the content based on viewport size.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,9 @@ node_modules
 # Optional npm cache directory
 .npm
 
+# Project npm config file
+.npmrc
+
 # Build Data
 dist/*
 es/*
--- a/examples/ComponentToPrint/index.tsx
+++ b/examples/ComponentToPrint/index.tsx
@@ -56,9 +56,7 @@ export class ComponentToPrint extends React.PureComponent<Props, State> {
           rel="stylesheet"
           href="../disabled.css"
         />
-        <style type="text/css" media="print">{"\
-   @page {\ size: landscape;\ }\
-"}</style>
+        <style type="text/css" media="print">{"@page { size: landscape; }"}</style>
         <div className="flash" />
         <table className="testClass">
           <thead>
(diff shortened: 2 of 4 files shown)


nosinformatica:
@MatthewHerbst I still have this problem how can i fix?

MatthewHerbst:
Hi @nosinformatica could you provide more information please? What browser/version and what version of `react-to-print`? An example such as in Codesandbox that shows the issue would be ideal, thanks!

nosinformatica:
> Hi @nosinformatica could you provide more information please? What browser/version and what version of `react-to-print`? An example such as in Codesandbox that shows the issue would be ideal, thanks!

Hi, i made the sandbox: https://codesandbox.io/p/sandbox/wrong-viewport-react-to-print-sandbox-zs4fww?file=%2Fsrc%2FApp.js

surajnarayanan1:
Hi, I'm facing a similar issue with the view port. Has this been resolved?

MatthewHerbst:
@surajnarayanan1 any issues you are facing here are almost certainly related to your styles, and is unlikely to be a bug in react-to-print: we don't change the viewport or your styles. If you are using a component library many will change the viewport when printing. If you want to post more about your situation maybe we can try and help there

PR #679, opened 2024-01-04
impl(lazy-initialize): set content for handlePrint() and handleClick() via second optional argument using [content]

### LINKED ISSUE #652: [ENHANCEMENT]: Set content function lazily to the handlePrint function returned by the useReactToPrint() hook
This is a feature enhancement request and i hope you would consider its' merits fairly.

I would like to have the content function set after the `useReactToPrint()` hook is initialised can called like so:

```javascript
import { useRef, useMemo } from "react";
import { useReactToPrint } from "react-to-print";


/* @NOTE: `navigator.clipboard` is undefined in Safari 12.1.x as well as the earlier versions 
  of other browsers like Chrome (Webkit), Firefox, Edge (EdgeHTML) */
/* @CHECK: https://developer.mozilla.org/en-US/docs/Web/API/Clipboard#clipboard_availability */
import "clipboard-polyfill/overwrite-globals"; /* @_SHIM: # */

/**
 * copyTextFactory
 *
 * @param {void}
 * @returns {void}
 */
function copyTextFactory() {
  return (textString = "") => {
    /* @NOTE: `navigator.clipboard.writeText(...)` throws vague error in Safari 13.1.x (and above) even when called in a real user context */
    /* @CHECK: https://developer.apple.com/forums/thread/691873 */
    try {
      navigator.clipboard.writeText(textString);
    } catch(_) { }
  };
}

/**
 * printPageFactory
 *
 * @param {Function} printer
 * @returns {void}
 */
function printPageFactory(printer) {
  return (componentRef = null) => {
     if (componentRef === null) {
        window.print();
     } else {
       /* @TODO: I need `content` to optionally be passed to the `handlePrint` function */
       printer(null, () => componentRef.current)
     }
  };
}

/* @NOTE: A ReactJS hook to house command design pattern actions for copying, printing e.t.c on a React app */
export const useCommands = () => {
  /* @HINT: COMMAND PATTERN */

  const printer = useReactToPrint();
  const commands = useRef({
    copy: copyTextFactory(),
    print: printPageFactory(printer)
  }).current;

  return useMemo(
    () => ({
      hub: {
        execute(commandName = "", ...args) {
          if (typeof commands[commandName] === "function") {
            const commandRoutine = commands[commandName];
            return commandRoutine.apply(null, args);
          }
        }
      }
    }),
    /* eslint-disable-next-line react-hooks/exhaustive-deps */
    []
  );
};
```

I could ready a PR soon if you accept this enhancement request and you don't mind (since it's a small change)

isocroft:
Fixes #652

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -206,7 +206,13 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
         }
     }
 
-    public handleClick (event: unknown, content?: (() => React.ReactInstance | null)) {
+    
+    public handleClick (
+        /* eslint-disable-next-line no-unused-vars */
+        // @ts-expect-error variable is unused
+        event: unknown, 
+        content?: (() => React.ReactInstance | null)
+    ) {
         const {
             onBeforeGetContent,
             onPrintError,
@@ -228,8 +234,6 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
         } else {
             this.handlePrint(content);
         }
-
-        return Boolean(event), true;
     }
 
     public handlePrint = (optionalContent?: (() => React.ReactInstance | null)) => {
@@ -602,12 +606,7 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
             }
 
             const value = { 
-                handlePrint: (
-                    event: unknown,
-                    content?: (() => React.ReactInstance | null)
-                ) => {
-                    return this.handleClick(event, content);
-                }
+                handlePrint: this.handleClick.bind(this)
             };
 
             return (
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -5,13 +5,13 @@ const contextEnabled = Object.prototype.hasOwnProperty.call(React, "createContex
 const hooksEnabled = Object.prototype.hasOwnProperty.call(React, "useMemo") && Object.prototype.hasOwnProperty.call(React, "useCallback");
 
 export interface IPrintContextProps {
-    handlePrint: (event: React.MouseEvent, content?: (() => React.ReactInstance | null)) => void,
+    handlePrint: (event: unknown, content?: (() => React.ReactInstance | null)) => void,
 }
 const PrintContext = contextEnabled ? React.createContext({} as IPrintContextProps) : null;
 export const PrintContextConsumer = PrintContext ? PrintContext.Consumer : () => null;
 
 export interface ITriggerProps<T> {
-    onClick: (event: React.MouseEvent) => void;
+    onClick: (event: unknown) => void;
     ref: (v: T) => void;
 }
 
@@ -206,7 +206,7 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
         }
     }
 
-    public handleClick (event: React.MouseEvent, content?: (() => React.ReactInstance | null)) {
+    public handleClick (event: unknown, content?: (() => React.ReactInstance | null)) {
         const {
             onBeforeGetContent,
             onPrintError,
@@ -229,7 +229,7 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
             this.handlePrint(content);
         }
 
-        return Boolean(event);
+        return Boolean(event), true;
     }
 
     public handlePrint = (optionalContent?: (() => React.ReactInstance | null)) => {
@@ -603,8 +603,8 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
 
             const value = { 
                 handlePrint: (
-                    event: React.MouseEvent,
-                    content: (() => React.ReactInstance | null)
+                    event: unknown,
+                    content?: (() => React.ReactInstance | null)
                 ) => {
                     return this.handleClick(event, content);
                 }
@@ -620,7 +620,7 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
 }
 
 type UseReactToPrintHookReturn = (
-    event: React.MouseEvent,
+    event: unknown,
     content?: (() => React.ReactInstance | null)
 ) => void;
 
@@ -642,7 +642,7 @@ export const useReactToPrint = (props: IReactToPrintProps): UseReactToPrintHookR
     );
 
     return React.useCallback(
-        (event: React.MouseEvent, content?: (() => React.ReactInstance | null)) => {
+        (event: unknown, content?: (() => React.ReactInstance | null)) => {
         /* eslint-disable-next-line @typescript-eslint/unbound-method */
         const triggerPrint = wrapCallbackWithArgs(reactToPrint, reactToPrint.handleClick, content);
         // NOTE: `event` is a no-use argument
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -210,7 +210,7 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
         }
     }
 
-    public handlePrint = () => {
+    public handlePrint = (lazyOption?: { content: (() => React.ReactInstance) }) => {
         const {
             bodyClass,
             content,
@@ -220,7 +220,7 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
             nonce,
         } = this.props;
 
-        const contentEl = content();
+        const contentEl = lazyOption && typeof lazyOption.content === "function" ? lazyOption.content() : content();
 
         if (contentEl === undefined) {
             this.logMessages(['To print a functional component ensure it is wrapped with `React.forwardRef`, and ensure the forwarded ref is used. See the README for an example: https://github.com/gregnb/react-to-print#examples']);
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -212,25 +212,24 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
             onPrintError,
         } = this.props;
 
-        // NOTE: `event` is a no-use argument (necessary for backward compatibility with older versions)
-        const __handlePrint = wrapCallbackWithArgs(this, this.handlePrint, event);
-
         if (onBeforeGetContent) {
             const onBeforeGetContentOutput = onBeforeGetContent();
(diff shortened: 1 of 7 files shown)


MatthewHerbst:
Thanks! Will take a look at this later this evening

isocroft:
Hello @MatthewHerbst , i was hoping to get some feedback on this as soon as you can be available say this weekend ?

MatthewHerbst:
Hi, apologies, have been slammed with work. Promise to review this within the next 48 hours

src/index.tsx
```diff
@@ -25,6 +25,12 @@ type Font = {
 
 type PropertyFunction<T> = () => T;
 
+function partialEnd<R, O>(thisArg: O, callback: (...curryArgs: any[]) => R, ...predefinedArgs: unknown[]) {
+    return function (...args: unknown[]) {
+      return callback.apply(thisArg, [...args, ...predefinedArgs]);
+    };
+}
```
MatthewHerbst: I think this is the biggest part of the PR I'm having trouble with. I understand what it's doing, but  it seems overly complex? Or maybe I'm just wishful thinking

src/index.tsx
```diff
@@ -5,13 +5,13 @@ const contextEnabled = Object.prototype.hasOwnProperty.call(React, "createContex
 const hooksEnabled = Object.prototype.hasOwnProperty.call(React, "useMemo") && Object.prototype.hasOwnProperty.call(React, "useCallback");
 
 export interface IPrintContextProps {
-    handlePrint: () => void,
+    handlePrint: (event: React.MouseEvent, lazyOption?: { content: (() => React.ReactInstance) }) => void,
```
MatthewHerbst: Any thoughts on just passing the content as the arg rather than using the object?
```suggestion
    handlePrint: (event: React.MouseEvent, content?: () => React.ReactInstance) => void,
```

src/index.tsx
```diff
@@ -5,13 +5,13 @@ const contextEnabled = Object.prototype.hasOwnProperty.call(React, "createContex
...
 }
 const PrintContext = contextEnabled ? React.createContext({} as IPrintContextProps) : null;
 export const PrintContextConsumer = PrintContext ? PrintContext.Consumer : () => null;
 
 export interface ITriggerProps<T> {
-    onClick: () => void;
+    onClick: (event?: React.MouseEvent) => void;
```
MatthewHerbst: When would we not pass an event?

src/index.tsx
```diff
@@ -220,7 +229,13 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
             nonce,
         } = this.props;
 
-        const contentEl = content();
+        let contentEl = lazyOption && typeof lazyOption.content === "function" ? lazyOption.content() : null;
```
MatthewHerbst:
```suggestion
        let contentEl = typeof lazyOption?.content === "function" ? lazyOption.content() : null;
```

src/index.tsx
```diff
@@ -220,7 +229,13 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
...
+        
+        if (!contentEl) {
+            if (typeof content === "function") {
+                contentEl = content();
+            }
+        }
```
MatthewHerbst: I believe `content` is already typed as a function at this point?
```suggestion
        if (!contentEl && content) {
                contentEl = content();
        }
```

MatthewHerbst (commented):
Looks good, just curious if there's a way to simplify the logic at all maybe?

src/index.tsx
isocroft: This looks good @MatthewHerbst ! Thanks for the suggestion (collapsing into a single `if`). I will apply it

src/index.tsx
isocroft: My bad, there is no time when the `event` object need be optional.

Thanks for catching that

src/index.tsx
isocroft: That's okay, nice suggestion will do so.

src/index.tsx
isocroft: Yeah, you are right. I think this is because of the naming of the generic types and the name of the function itself. This function is important to the backwards compatibility so one does not need to always pass the event object when calling `handlePrint()` returned from `useReactPrint({ ... })`.

Without currying, this will not work as expected and the next release will have to be a major version number bump release as there will be breaking changes.

I have taken the time to rename and format the function and the generic typing accordingly

src/index.tsx
isocroft: This is a very good suggestion. I will modify accordingly

isocroft:
Hello. @MatthewHerbst , I am done with the change requests you mandated based on your review.

Thanks a lot 🙏🏾

MatthewHerbst (commented):
Looks great! One last request: can we make the example much, much shorter? The examples for the other uses are about 25 lines each, and don't reference any random external dependencies. Can we make the added example similarly short, and remove any reference to external dependencies that most readers of the README will not be familiar with?

isocroft:
Ok @MatthewHerbst , I will do just that.

Thanks

isocroft:
All done now @MatthewHerbst 👍🏾

src/index.tsx
```diff
@@ -5,13 +5,13 @@ const contextEnabled = Object.prototype.hasOwnProperty.call(React, "createContex
 const hooksEnabled = Object.prototype.hasOwnProperty.call(React, "useMemo") && Object.prototype.hasOwnProperty.call(React, "useCallback");
 
 export interface IPrintContextProps {
-    handlePrint: () => void,
+    handlePrint: (event: React.MouseEvent, content?: (() => React.ReactInstance | null)) => void,
```
MatthewHerbst: I realized one last thing: it's possible that some users pass this function to other type of event callbacks, and this type would break there. For example:

Screenshot 2024-02-07 001840

Can we change this to just be `unknown` for now since we don't need it?
```suggestion
    handlePrint: (event: unknown, content?: (() => React.ReactInstance | null)) => void,
```

MatthewHerbst: I think we'll need this change in all the places we've typed this, to maintain backwards compatibility

MatthewHerbst (commented):
Ah, realized one last thing! Otherwise looks fantastic, and thank you so much for adding the examples!

src/index.tsx
isocroft: All done now @MatthewHerbst . Thanks a lot

src/index.tsx
```diff
@@ -196,21 +216,23 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
...
             } else {
-                this.handlePrint();
+                this.handlePrint(content);
             }
         } else {
-            this.handlePrint();
+            this.handlePrint(content);
         }
+
+        return Boolean(event), true;
```
MatthewHerbst: What is the purpose of this return statement? Can it be removed? It seems to always return `true`, even if `Boolean(event) === false` (honestly when I first saw this I didn't even think it was valid JS, learn something new every day!

src/index.tsx
```diff
@@ -575,7 +601,14 @@ export default class ReactToPrint extends React.Component<IReactToPrintProps> {
...
             }
 
-            const value = { handlePrint: this.handleClick };
+            const value = { 
+                handlePrint: (
+                    event: unknown,
+                    content?: (() => React.ReactInstance | null)
+                ) => {
+                    return this.handleClick(event, content);
+                }
```
MatthewHerbst: Looking at this, how is it different from what we had before? Why is `{ handlePrint: this.handleClick }` still not good enough? Were the types having issues?

MatthewHerbst (commented):
Nice! Happy with this! Two small nits before we merge

src/index.tsx
isocroft: You're right. `{ handlePrint: this.handleClick.bind(this) }` is good enough. I have modified this accordingly

src/index.tsx
isocroft: There's a need for this line else `event` becomes and _unused variable_ and the `tsc` starts throwing errors

MatthewHerbst: Does changing the parameter name to `_event` solve that?

isocroft: However, i have resolved it using an `eslint` comment

isocroft:
Hello @MatthewHerbst , all done here

MatthewHerbst (approved):

MatthewHerbst:
Published in v2.15.0!

Bowbaq:
This seems to break existing code by adding this new mandatory `event` argument, which doesn't appear to be used for anything.

```
> tsc && vite build

src/components/print-table/index.tsx:23:5 - error TS2554: Expected 1-2 arguments, but got 0.

23     handlePrint();
       ~~~~~~~~~~~~~

  node_modules/react-to-print/lib/hooks/useReactToPrint.d.ts:3:35
    3 type UseReactToPrintHookReturn = (event: unknown, content?: (() => React.ReactInstance | null)) => void;
                                        ~~~~~~~~~~~~~~
    An argument for 'event' was not provided.


Found 1 error in src/components/print-table/index.tsx:23
```

MatthewHerbst:
> This seems to break existing code by adding this new mandatory `event` argument, which doesn't appear to be used for anything.
> 
> ```
> > tsc && vite build
> 
> src/components/print-table/index.tsx:23:5 - error TS2554: Expected 1-2 arguments, but got 0.
> 
> 23     handlePrint();
>        ~~~~~~~~~~~~~
> 
>   node_modules/react-to-print/lib/hooks/useReactToPrint.d.ts:3:35
>     3 type UseReactToPrintHookReturn = (event: unknown, content?: (() => React.ReactInstance | null)) => void;
>                                         ~~~~~~~~~~~~~~
>     An argument for 'event' was not provided.
> 
> 
> Found 1 error in src/components/print-table/index.tsx:23
> ```

Ah crap, you're totally right, this should be optional. Will push out a fix now

MatthewHerbst:
@Bowbaq fix published in v2.15.1, please let me know if you run into any additional issues, and thank you so much for the report!

PR #723, opened 2024-08-14
Implement a new option that adds support for copying shadow root content

### LINKED ISSUE #176: React component wrapped inside custom element (web component)
Thank you for react-to-print. Excellent job! In my application users can toggle between a dark and a light (Bootstrap-based) theme. For when the dark theme is active, I figure I "wrap" my React component inside a custom element (web component) to take advantage of the shadow DOM. That would allow me to switch to light theme just for printing purposes without affecting the dark theme that the user is looking at.

When I use react-to-print to print this React class component directly it works as advertised. When I wrap it inside a custom element, everything seem to go through the motions but nothing is rendered. The react-to-print callbacks are called and the React component's render method is called. When using the custom element directly inside another React component (no printing), it renders to the screen just fine.

Would you have any idea what might be going on here?

boehlke:
Resolves #176


src/hooks/useReactToPrint.ts
```diff
@@ -1,6 +1,6 @@
 import { useCallback } from "react";
 
-import { Font } from "../types/Font";
+import { Font } from "../types/font";
```
MatthewHerbst, line 3: Is this change intentional?

src/types/UseReactToPrintOptions.ts
```diff
@@ -1,5 +1,5 @@
 import { RefObject } from "react";
-import type { Font } from "./Font";
+import type { Font } from "./font";
```
MatthewHerbst, line 2: Is this change intentional?

src/types/UseReactToPrintOptions.ts
```diff
@@ -54,4 +54,6 @@ export interface UseReactToPrintOptions {
     print?: (target: HTMLIFrameElement) => Promise<any>;
     /** When passed, prevents `console` logging of errors */
     suppressErrors?: boolean;
-}
\ No newline at end of file
+    /** When passed, shadow root content will be copied */
+    copyShadowRoots?: boolean;
```
MatthewHerbst, line 59: Can we add this to the README as well please?

src/utils/clone.ts
```diff
@@ -0,0 +1,31 @@
+export function cloneShadowRoots(sourceNode: Node, targetNode: Node) {
+    if(sourceNode instanceof Element && targetNode instanceof Element) {
```
MatthewHerbst:
```suggestion
    if (sourceNode instanceof Element && targetNode instanceof Element) {
```

src/utils/handlePrintWindowOnLoad.ts
```diff
@@ -1,7 +1,8 @@
 import { logMessages } from "./logMessage";
 import { startPrint } from "./startPrint";
-import { Font } from "../types/Font";
+import { Font } from "../types/font";
```
MatthewHerbst: Is this change intentioinal?

src/utils/handlePrintWindowOnLoad.ts
```diff
@@ -61,7 +62,10 @@ export function handlePrintWindowOnLoad(
     const domDoc = printWindow.contentDocument || printWindow.contentWindow?.document;
 
     if (domDoc) {
-        domDoc.body.appendChild(clonedContentNode);
+        const appendedContentNode = domDoc.body.appendChild(clonedContentNode);
+        if(options.copyShadowRoots) {
```
MatthewHerbst:
```suggestion
        if (options.copyShadowRoots) {
```

MatthewHerbst: Nit: can you please destructure `copyShadowRoots` above like on lines 49-56

src/utils/clone.ts
```diff
@@ -0,0 +1,31 @@
...
+        const shadowRoot = sourceNode.shadowRoot;
+        if (shadowRoot !== null) {
+            // Create a new Shadow Root for the target element
+            const copiedShadowRoot = targetNode.attachShadow({mode: shadowRoot.mode});
```
MatthewHerbst:
```suggestion
            const copiedShadowRoot = targetNode.attachShadow({ mode: shadowRoot.mode. });
```

src/utils/clone.ts
```diff
@@ -0,0 +1,31 @@
...
+
+            // Clone the contents of the original ShadowRoot manually
+            const originalShadowContent = shadowRoot.innerHTML;
+
+            // Create a temporary container to hold the content to copy
+            const tempContainer = targetNode.ownerDocument.createElement(shadowRoot.children[0].tagName);
```
MatthewHerbst: Will the shadow root always have at least one child?

src/utils/clone.ts
```diff
@@ -0,0 +1,31 @@
...
+            tempContainer.innerHTML = originalShadowContent;
+
+            // Clone and append the Shadow DOM content
+            copiedShadowRoot.appendChild(tempContainer);
+
+            // Recursively clone any nested Shadow DOMs within this Shadow DOM content
```
MatthewHerbst: This recursion and the one below is what scared me from doing this in the past, I'm worried about performance problems on pages with large amounts of shadow content. Maybe we worry about that if/when it happens though

MatthewHerbst (commented):
This looks great, thanks! Just a few things to cleanup please, mostly we need to add the docs for it to the README

src/types/UseReactToPrintOptions.ts
boehlke: There is only a "font.ts", no "Font.ts". My build failed with 'from "./Font"...

src/utils/handlePrintWindowOnLoad.ts
boehlke: Yes... like pointed out above.

src/utils/clone.ts
boehlke: What do think about a warning in the README to address this issue.
I optimized/simplified the clone routine, treewalker should be quite efficient. Please have a look.

src/utils/clone.ts
boehlke: Simplified that... the old code version was a bit hacky. Much cleaner now.

src/hooks/useReactToPrint.ts
boehlke: Can you resolve this if lower case works fine for you as well? Do you work with NTFS? For case sensitive FS, this should lead to an error.

src/types/UseReactToPrintOptions.ts
MatthewHerbst: I think my computer is lying to me!

Screenshot 2024-08-19 at 1 51 58 AM
Screenshot 2024-08-19 at 1 52 09 AM

src/utils/clone.ts
MatthewHerbst: I like that, thanks!

src/utils/clone.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    const sourceElements = collectElements(sourceNode as HTMLElement);
+    const targetElements = collectElements(targetNode as HTMLElement);
+
+    for (let i = 0; i < sourceElements.length; i++) {
+        const sourceElement = sourceElements[i];
+        const targetElement = targetElements[i];
+
+        const shadowRoot = sourceElement.shadowRoot;
+        if (shadowRoot !== null) {
```
MatthewHerbst, line 34: Should we guard that `targetElement` is also defined, since it's possible `targetElements` may have a different length than `sourceElements` (this would likely only happen if the user isn't waiting for their DOM to be settled before printing, but if they make a mistake there this will hard crash them. I just added a similar check when iterating over `canvas` elements): https://github.com/MatthewHerbst/react-to-print/blob/8188f8406ca17752a513b5bf46172cf953df3805/src/utils/handlePrintWindowOnLoad.ts#L121-L127

src/utils/clone.ts
MatthewHerbst: `TreeWalker` is nice!

src/utils/clone.ts
boehlke: Please review, I added a similar error handling like you suggested.

MatthewHerbst: Perfect, thanks!

MatthewHerbst (approved):
Thanks so much for this! I'll get it into a `beta-3` release this evening 🥳

MatthewHerbst:
Published in `v3.0.0-beta-3` 🥳

PR #732, opened 2024-09-03
fix: Fix readme example

sadeghpro:


MatthewHerbst (changes requested):
Hey! Really appreciate you taking the time to make a PR to try and help out!

The README example is actually correct, when looking at the `master` branch. I'm currently working on v3 of this library, with `v3.0.0-beta-3` being the most recently tagged beta version published on `npm`. Until it moves from beta to production, npm will continue showing the `2.15.1` README and that will be the default version you install (via `npm i react-to-print`) unless you explicitly request the beta version (`npm i react-to-print@3.0.0-beta-3`).

If you want to see version of the code corresponding with v2.15.1 on GitHub you can use the tags selector to choose that version:

Screenshot 2024-09-03 at 10 52 49 AM

Issue #187, opened 2019-12-09
Printing entire UI on mobile


hrutvikpatel:
ReacToPrint is able to print the specific react component I want to print on browsers on the desktop and on the IPad safari works as well. However, when printing the react component using chrome on the IPad the entire UI is printed.

I checked the console and there were no errors present.

I also checked that the ref returned to the content property is referring to the react component that needs to be printed.

I am not sure why ReactToPrint defaults to printing the entire UI on chrome on the IPad, I was wondering if someone else also ran into this issue.

hrutvikpatel:
I am trying to understand the code and was wondering what the purpose of the setTimeout is in the startPrint method?

hrutvikpatel:
It appears to be an issue with iframe. Where `iframe.contentwindow.print()` seems to have an issue printing on safari/chrome on IOS. It prints the entire UI instead. I wonder if there is a workaround to this.

MatthewHerbst:
Unfortunately we have no control (to my knowledge) of what the browser decides to do after we call `.print()`. It sounds like those browsers print everything open, not just the iframe. If you know of a solution we can use to fix this issue I would be happy to include it.

ekediala:
> ReacToPrint is able to print the specific react component I want to print on browsers on the desktop and on the IPad safari works as well. However, when printing the react component using chrome on the IPad the entire UI is printed.
> 
> I checked the console and there were no errors present.
> 
> I also checked that the ref returned to the content property is referring to the react component that needs to be printed.
> 
> I am not sure why ReactToPrint defaults to printing the entire UI on chrome on the IPad, I was wondering if someone else also ran into this issue.

Hi. Ran into the same issue. Did you find a workaround?

Cheers.

MatthewHerbst:
@ekediala as noted above, to the best of my knowledge this is not something we can fix. If you know of a solution please let me know

ekediala:
@MatthewHerbst thank you. Will let you know if I do.

nate-vukovich:
Any news on this one?

Also ran into this when building a GatsbyJS website and I found the package perfect for the job.

Will also do some digging if there is something doable here ;)

joungwoo-jay:
Hmm looking for any news for this issue. 
We have tested all other browsers with this great library, but I we were not able to print in Chrome on iPad.
The print preview returns entire application... hmm, I will try to use different print method, and update here.

james-e-morris:
[...]

PR #764, opened 2024-11-24
Implement workaround for print iframe removal on focus event

zainphp:
This pull request introduces an important update to the `startPrint` function in the `src/utils/startPrint.ts` file. The update implements a workaround to address the issue where the print iframe is removed immediately after the print dialog closes. This fixes the problem described in issue #187, where the iframe was being removed too early—before the print dialog appeared—causing the print preview to display the entire UI instead of the intended content.

* [`src/utils/startPrint.ts`](diffhunk://#diff-f9538760ee08f21e6507942479209741c43ae72102c42f8f6d4acfe09c5eed93R78-R90): Added a `focus` event listener to remove the print iframe after the print dialog closes, addressing the lack of a reliable `afterprint` event in most browsers.

### DIFF
--- a/src/utils/startPrint.ts
+++ b/src/utils/startPrint.ts
@@ -75,8 +75,19 @@ export function startPrint(printWindow: HTMLIFrameElement, options: UseReactToPr
                     });
                 }
 
-                onAfterPrint?.();
-                removePrintIframe(preserveAfterPrint);
+                /**
+				 * Remove the print iframe after the print dialog closes that determined by the main
+				 * window is regaining focus after the print dialog closes. This workaround is applied
+				 * due to the lack of a reliable `afterprint` event in most browsers.
+				 */
+				window.addEventListener(
+					"focus",
+					() => {
+						onAfterPrint?.();
+						removePrintIframe(preserveAfterPrint);
+					},
+					{ once: true }
+				);
             }
         } else {
             logMessages({


MatthewHerbst:
Hey @sensasi-delight, thanks so much for this! I've been trying to think of situations where this might not work, or if it might cause problems for anything. Going to think about it for a few more days and then will merge if nothing comes up. Really appreciate you looking into this!

zainphp:
Thank you for the response! Please let me know if there's anything I can clarify or improve further. I'm happy to help.

jeremycharron:
Hello @MatthewHerbst , I was wondering if you were planning to merge this (as a beta version)?
I was considering doing a patch in my project, but I'd rather wait an official tag.

MatthewHerbst:
Hey! Apologies this has been delayed, I've been trying to find ways to break it. It seems like a simple change but event handlers have all sorts of weird behaviors so there's actually a lot of room for new edge cases to be introduced here. I will merge this and get it released as a beta in the next few hours

MatthewHerbst:
I'm particularly concerned about the `{ once: true }` so have been looking for ways to mess around with that

MatthewHerbst:
`3.0.3-beta-1` has been published, please let me know if you encounter any issues

MatthewHerbst:
@sensasi-delight this seems to have caused a new issue, https://github.com/MatthewHerbst/react-to-print/issues/779, where browsers that don't automatically return focus to the page will never have `onAfterPrint` called. I looked into the [HTML `window.print` spec](https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#printing) and there is no requirement to return focus to the page from what I can find, though not doing so seems like very poor UX. Curious for your thoughts here.

zainphp:
Dear @MatthewHerbst,

First of all, I want to apologize for the unintended issue caused by this pull request. It did resolve the problem on Chrome for Android on my end, but I didn't test it on Vivaldi, and it’s entirely possible that Vivaldi behaves differently. I completely agree that the lack of focus return is poor UX.

I’ve been thinking of a few alternative workarounds, such as leveraging the `beforeunload` event on the print window or using the `visibilitychange` event on the main window. I’ll experiment with these approaches and, if you’re open to it, I’d be happy to submit a new pull request once I find a reliable solution.

MatthewHerbst:
Hey, all good, it happens! Appreciate you responding so quick! I'll take a look at the new PR in a bit, thank you!

PR #772, opened 2024-12-16
execution exception for empty video src

Can-Chen:
when using vidstack/player to play videos, if the area where the video is rendered is not scrolled, the video src will be empty at this time. After calling reactToPrint, the browser's printing function will not be invoked and no errors will be thrown.

### DIFF
--- a/src/utils/handlePrintWindowOnLoad.ts
+++ b/src/utils/handlePrintWindowOnLoad.ts
@@ -188,7 +188,9 @@ export function handlePrintWindowOnLoad(
             } else {
                 if (videoNode.readyState >= 2) { // Check if the video has already loaded a frame
                     markLoaded(videoNode);
-                } else {
+                } else if (!videoNode.src) {
+                    markLoaded(videoNode, ["video src empty", videoNode, "Error"]);
+                } else{
                     videoNode.onloadeddata = () => {
                         markLoaded(videoNode)
                     };


MatthewHerbst:
Hi @Can-Chen, apologies for the delayed response here. I just tried testing this out with the local example and was not able to reproduce it. Are you able to share a [Codesandbox](https://codesandbox.io/p/sandbox/rzdhd) or similar where I can see this issue?

Can-Chen:
[react-test.zip](https://github.com/user-attachments/files/18224531/react-test.zip)

problems will occur when the src of the video element is empty.

Can-Chen:
> [react-test.zip](https://github.com/user-attachments/files/18224531/react-test.zip)

problems will occur when the src of the <video> element is empty.

MatthewHerbst:
I'm not able to download files right now on the computer I have access to; could you please either  use Codesandbox or post the code in the comment? Thank you

Can-Chen:
https://codesandbox.io/p/sandbox/react-to-print-test-39cm9v

MatthewHerbst:
Thanks! When I go to that example I do not see a problem. I am using Chrome 131. What browser are you using?


https://github.com/user-attachments/assets/73f73fd7-d34b-4cc5-9e8e-23ea591ddea1

Can-Chen:
after the page is rendered, the video element will trigger when it has never appeared in the viewport.

https://github.com/user-attachments/assets/be4d724a-e478-4cea-90f3-1300a201e769

MatthewHerbst:
Ah, I see now, thank you!

MatthewHerbst:
I wonder if there is a way to make it work instead of having an error; I will look for a solution. If I cannot find one I will merge this PR

Can-Chen:
ok

MatthewHerbst:
Going to merge this while I look for a fix, I think this is good protection to have either way. Thank you!!

MatthewHerbst:
I do wonder if this is an issue with `MediaPlayer`. I am not able to replicate this issue with the video in the react-to-print examples: https://codesandbox.io/p/sandbox/rzdhd

MatthewHerbst:
Yeah, I believe the root issue is with the `MediaPlayer`. I filed an issue here: https://github.com/vidstack/player/issues/1548
This check is still good to have though to be safe against bad `video` node constructions, so thank you, I will release a new version with this fix shortly.

Can-Chen:
> Yeah, I believe the root issue is with the `MediaPlayer`. I filed an issue here: [vidstack/player#1548](https://github.com/vidstack/player/issues/1548) This check is still good to have though to be safe against bad `video` node constructions, so thank you, I will release a new version with this fix shortly.

The properties "load" and "posterLoad" of the `MediaPlayer` need to be set to "eager".

https://vidstack.io/docs/player/core-concepts/loading/?styling=default-theme#load-strategies

MatthewHerbst:
Yeah, re-reading [the spec](https://html.spec.whatwg.org/multipage/media.html#the-video-element), `src` is definitely optional in many scenarios. No need to try and "fix" it, this check you've added is enough 🥳

MatthewHerbst:
Published the fix in version `3.0.4`. Thanks so much for the help @Can-Chen!

PR #780, opened 2025-01-07
Improve after print handling

### LINKED ISSUE #779: In 3.0.3 and later onAfterPrint is called only after clicking back on the page
When using version `3.0.3` or `3.0.4`, `onAfterPrint` is called only after clicking back on the page after print dialogue is closed.
This behavior is not present in `3.0.2`

zainphp:
This pull request resolves #779, which was caused by changes in #764, by separating the logic for mobile and desktop browsers. Additionally, I’ve improved the solution from PR #764 by using `setTimeout` instead of `addEventListener` to ensure proper execution after printing on mobile browsers.

I’ve also refactored repetitive code to make it cleaner and more maintainable.

Tested Successfully On:
```
 - Windows 11:
   - Chrome v131
   - Edge v131
   - Vivaldi v6
   - Vivaldi v7
- Android:
  - Chrome v131
  ```

### DIFF
--- a/src/utils/startPrint.ts
+++ b/src/utils/startPrint.ts
@@ -24,12 +24,14 @@ export function startPrint(printWindow: HTMLIFrameElement, options: UseReactToPr
         if (printWindow.contentWindow) {
             printWindow.contentWindow.focus(); // Needed for IE 11
 
+            function handleAfterPrint() {
+                onAfterPrint?.()
+                removePrintIframe(preserveAfterPrint)
+            }
+
             if (print) {
                 print(printWindow)
-                    .then(() => {
-                        onAfterPrint?.();
-                        removePrintIframe(preserveAfterPrint);
-                    })
+                    .then(handleAfterPrint)
                     .catch((error: unknown) => {
                         if (onPrintError) {
                             onPrintError('print', getErrorFromUnknown(error));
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/utils/startPrint.ts [hunk 1]
```diff
@@ -103,4 +103,39 @@ export function startPrint(printWindow: HTMLIFrameElement, options: UseReactToPr
...
+/**
+ * Determines if the current browser is a mobile browser by checking the `navigator.userAgent` 
+ * against a predefined list of common mobile browser identifiers.
+ * 
+ * Note: This function is not exhaustive and may not detect all mobile browsers, 
+ * as it is designed to remain simple and lightweight.
+ * 
+ * @see [Stack Overflow Reference](https://stackoverflow.com/a/11381730/4899926)
+ */
+function isMobileBrowser() {
```
MatthewHerbst, line 117: I've been extremely hesitant in the past to do any sort of browser detection, and have generally preferred to report upstream bugs and then shrug if they don't get fixed. Browser detection in one place often leads to a slippery slope of trying to fix things on a browser-by-browser basis. In this case, the spec is not clear, and so it's not clear that Vivaldi is actually doing anything wrong, by the letter at least. 

Maybe we should just bite the bullet and start doing this though 🤷

MatthewHerbst (commented):
I've filed an upstream bug report to Vivaldi (VB-112980). If we don't hear from them in the next few days I'll merge this.

src/utils/startPrint.ts
zainphp: Thank you for your response. I have already tried various other browser events, but unfortunately, they did not yield the desired results due to the behavioral differences between mobile and desktop browsers. As a result, I have no choice but to implement mobile browser detection to ensure they are handled differently.

MatthewHerbst (approved):

Issue #784, opened 2025-02-03
Tailwind css styling is not applied in PDF generate


joyal007:
When using Tailwind CSS with the configuration `important: "#body"`, components styled with Tailwind utilities fail to render correctly in PDFs generated using the react-to-print library. This issue arises because the dynamically created iframe used for printing does not include an element with the id="body".

```js
module.exports = {
  important: '#body',
  // other configurations
};

```

PR #785, opened 2025-02-03
Feat: Add custom id to Iframe body

joyal007:
Added optional argument `iframeBodyId` to useReactToPrint which will set id from iframe body.

This is potential fix for #784


MatthewHerbst:
Thanks, I'll play around with this later tonight

MatthewHerbst:
Is the goal to actually style the [body](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/body) of the document, or do you just need an element with an id of `"body"` to exist? For example, if you wrapped your content being printed in a `<div id="body">` does that solve the issue?

joyal007:
Yes it's also a fix, but i felt it's little inconvenient naming `id='body'` for a div + i need to do this for all the component that i wish to generate pdf.

Also we already have an body element with id='body'.

MatthewHerbst:
Hey, I haven't forgotten this, just been trying to think about the best path forward. We already have a `bodyClass` prop so this isn't unprecedented. My thinking is that instead of continually adding new props to modify the iframe, instead, I'm going to look at adding a callback function that exposes the iframe to the function, so any changes to it can be made prior to printing. Still thinking through that a little, but I will try to get to it soon

PR #831, opened 2026-01-28
Dual CJS/ESM builds (with vite)

### LINKED ISSUE #760: No ESM build?
Is there no ESM build? I'm trying to use it in Remix + Vite and Vite complains that it is a CJS module that cannot be imported.

andrew8er:
This PR replaces webpack for vite, which makes it easy to build both CJS and ESM. The dev workflow should be identical though. 

If you think that changing the build system is too much of a change, feel free to close this PR.

Fixes #760


vite.config.ts
```diff
@@ -0,0 +1,17 @@
+import react from '@vitejs/plugin-react'
+import { resolve } from 'path'
+import dts from 'unplugin-dts/vite'
```
MatthewHerbst, line 3: This relatively new and still in-beta package is my only concern with the PR. Is there another way to handle this?

MatthewHerbst (commented):
Amazing! I've been meaning to take a look at this for a long time, thank you for making the PR. I'll look to test it over the next day or two. I have one concern with one of the packages used, but otherwise looks great!

vite.config.ts
andrew8er: As far as I understand, this plugin is a port from [vite-plugin-dts](https://www.npmjs.com/package/vite-plugin-dts), which is already > 5 years old, and is now the default version. The old version is still available, but not recommended, according to the [Readme](https://github.com/qmhc/unplugin-dts#readme).

The resulting `.d.ts` files are identical, apart from using single quotes for import paths. I'm pretty confident that this will work.

I can refactor this PR to use the original plugin though, I don't see any downsides here.

MatthewHerbst: Hmm, thanks for the context, very helpful! Let me dig into it a little - I'm super conscious these days of pulling in new dependencies because of all the security concerns the ecosystem has been facing, but agree the original seems pretty highly trafficked

andrew8er: Very reasonable…

MatthewHerbst: Figured out how to do this. I couldn't push to your branch since it's off your fork: https://github.com/MatthewHerbst/react-to-print/pull/836/changes/500b33f3722596f23f6a47c3cc9ce789f3f8feb4

Made a new PR with it: https://github.com/MatthewHerbst/react-to-print/pull/836

andrew8er:
Closed in favor of #836

MatthewHerbst:
Published with `v3.3.0`, thank you!

andrew8er:
Thanks!

PR #833, opened 2026-02-03
feat: Support dynamic documentTitle via function callback

### LINKED ISSUE #832: [Feature Request] Support dynamic documentTitle via function callback
## Problem

Currently, `documentTitle` is captured at hook initialization time, making it impossible to set dynamic filenames (e.g., with timestamps) without recreating the hook.

## Use Case
Users need to include timestamps or other dynamic data in the print filename:
- Invoice with current date: `Invoice_2026-02-02.pdf`
- Report with generation time: `Report_14-30-25.pdf`
- B/L draft with timestamp: `BILL_OF_LADING_DRAFT_2026-02-02_09-30.pdf`
 

## Current Workaround
Users must either:
1. Accept static filenames (loses timestamp info)
2. Recreate the hook on every state change (performance issue)
3. Violate React Hooks rules by conditionally calling hooks
4. 
## Proposed Solution
Support function callbacks for `documentTitle`:
// Option 1: Allow function type
documentTitle?: string | (() => string);

// Option 2: Add new parameter
getDocumentTitle?: () => string;

Implementation
I'm willing to submit a PR for this feature. Would the maintainers be open to:
- [x] Option 1 (extend existing parameter)
- [ ] Option 2 (add new parameter)
- [ ] Alternative approach?

Henry-G-H-Huang:
Closes https://github.com/MatthewHerbst/react-to-print/issues/832

## Summary
This PR adds support for dynamic `documentTitle` by allowing it to accept a function callback in addition to the existing string type.
##  Motivation
Users need to include dynamic data (timestamps, user info, etc.) in print filenames. Currently, `documentTitle` is captured at hook initialization, making this impossible without recreating the hook.
 ## Changes
- ✅ Modified `UseReactToPrintOptions` type to support `string | (() => string)`
- ✅ Updated `useReactToPrint` hook to evaluate function callbacks at print time
- ✅ Updated `startPrint` to handle function type resolution
- ✅ Added `DynamicDocumentTitle` example demonstrating timestamp usage
- ✅ Updated README documentation with new usage examples
- ✅ Maintains 100% backward compatibility
 ## Usage Example
const printFn = useReactToPrint({
  contentRef: componentRef,
  documentTitle: () => `Invoice_${new Date().toISOString().split('T')[0]}`,
});
This generates filenames like Invoice_2026-02-03.pdf with the current date when printing.
## Testing
- ✅ All existing lint checks pass
- ✅ TypeScript type checking passes
- ✅ Verified backward compatibility with string type
- ✅ No breaking changes

### DIFF
--- a/src/utils/startPrint.ts
+++ b/src/utils/startPrint.ts
@@ -49,24 +49,28 @@ export function startPrint(printWindow: HTMLIFrameElement, options: UseReactToPr
                     const tempContentDocumentTitle = printWindow.contentDocument?.title ?? '';
                     const tempOwnerDocumentTitle = printWindow.ownerDocument.title;
 
+                    const resolvedTitle = typeof documentTitle === 'function' 
+                        ? documentTitle() 
+                        : documentTitle;
+
                     // Override page and various target content titles during print
                     // NOTE: some browsers seem to take the print title from the highest level
                     // title, while others take it from the lowest level title. So, we set the title
                     // in a few places and hope the current browser takes one of them :pray:
-                    if (documentTitle) {
+                    if (resolvedTitle) {
                         // Print filename in Chrome
-                        printWindow.ownerDocument.title = documentTitle;
+                        printWindow.ownerDocument.title = resolvedTitle;
 
                         // Print filename in Firefox, Safari
                         if (printWindow.contentDocument) {
-                            printWindow.contentDocument.title = documentTitle;
+                            printWindow.contentDocument.title = resolvedTitle;
                         }
                     }
 
                     printWindow.contentWindow.print();
 
                     // Restore the page's original title information
-                    if (documentTitle) {
+                    if (resolvedTitle) {
                         printWindow.ownerDocument.title = tempOwnerDocumentTitle;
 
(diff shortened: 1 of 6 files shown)


src/utils/startPrint.ts
```diff
@@ -49,24 +49,28 @@ export function startPrint(printWindow: HTMLIFrameElement, options: UseReactToPr
                     const tempContentDocumentTitle = printWindow.contentDocument?.title ?? '';
                     const tempOwnerDocumentTitle = printWindow.ownerDocument.title;
 
+                    const resolvedTitle = typeof documentTitle === 'function'
```
MatthewHerbst, line 52: Do you need to do the check again here? I believe at this point it's for sure a `string` based on it being set as such when you did this same logic in the hook.

MatthewHerbst (commented):
Looks great! Just one question about needing to double-up on the check

MatthewHerbst:
Replaced this PR with https://github.com/MatthewHerbst/react-to-print/pull/837 which only does the check once

MatthewHerbst:
Published with `v3.3.0`, thank you!

